PR #2, opened 2021-03-14
style: Add support for applying a stylesheet to the dom

robinlinden:
@zer0-one PTAL. I've invited you to collaborate on this repo. We'll move it to a separate org later.

This handles taking the prepared css rules and applying them to the dom tree, returning an overlay tree where the nodes are decorated with the styles that matched.

I didn't add it in this PR, but this overlay tree should probably prune all branches with `display: none` set.

Next one will be the layout library that will consume the styled tree and use the height/width, padding, and margin attributes to create a skeleton with known positions and sizes for rendering.

### DIFF
--- a/dom/dom.h
+++ b/dom/dom.h
@@ -40,6 +40,22 @@ inline Node create_element_node(std::string_view name, AttrMap attrs, std::vecto
 
 std::string to_string(Node const &node);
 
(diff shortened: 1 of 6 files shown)


Zer0-One (approved):
My only nit is that `matching_rules()` is difficult to read, otherwise LGTM.

PR #7, opened 2021-04-04
Use system fmt instead of spdlog bundled version

Zer0-One:
CMake build fails if distro package does not ship with bundled fmt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,12 @@ find_package(Asio REQUIRED)
 find_package(fmt REQUIRED)
 find_package(spdlog REQUIRED)
 
+if(${CMAKE_PROJECT_NAME} STREQUAL ${PROJECT_NAME})
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY lib)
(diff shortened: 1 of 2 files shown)


robinlinden (changes requested):
You should probably add `find_package(fmt REQUIRED)` before `include(FindPackageHandleStandardArgs)` and add `fmt::fmt` to the `target_link_libraries` to make sure it's linked against too?

CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ find_package(Asio REQUIRED)
 find_package(fmt REQUIRED)
 find_package(spdlog REQUIRED)
 
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY lib)
```
robinlinden, line 11: I swear these weren't here the first review. :P
Please move to a separate commit, also move binaries to bin for consistency, and wrap in an if checking that this is the top-level project. I've more than once been annoyed that some random library I pull in via FetchContent randomly changed my output dirs.

robinlinden (changes requested):

robinlinden (approved):

PR #33, opened 2021-09-21
css: Add support for shorthand property padding

ml85:
Rules for expanding padding :
https://developer.mozilla.org/en-US/docs/Web/CSS/padding


robinlinden (approved):
Looks great! Tests and all. 😁 Some minor style misses, but I'll leave them since I've got clang-format pretty much done and that will involve a reformat of all my style misses already.

PR #36, opened 2021-09-23
Display the currently hovered element in the debug GUI

robinlinden:


browser/gui.cpp
```diff
@@ -91,20 +96,50 @@ int main(int argc, char **argv) {
...
+                            glTranslatef(0, translation, 0);
+                            scroll_offset_y -= translation;
                             break;
                         }
                         case sf::Keyboard::Key::K: {
-                            float scroll = event.key.shift ? 20.f : 5.f;
-                            glTranslatef(0, scroll, 0);
+                            float translation = event.key.shift ? 20.f : 5.f;
+                            glTranslatef(0, translation, 0);
+                            scroll_offset_y -= translation;
```
ml85, line 107: Some minor code duplication here since both cases are quite similar. However, it's ok for me.

ml85 (approved):

browser/gui.cpp
robinlinden: Yeah, the entire test GUI is a bit of a mess. I'll do some cleanup in a follow-up PR since it'll be quite a large (unrelated to this feature) change. Appreciate the review!

PR #39, opened 2021-09-26
Take padding into account when layouting

robinlinden:
Part of #20


mkiael (approved):
Looks solid!

robinlinden:
Thank you for the review!

PR #40, opened 2021-09-27
browser/gui: Give the application a class to live in

robinlinden:
@mkiael PTAL! I tried to not refactor things in this PR, but I still accidentally moved some things (e.g. the scroll code duplication you commented on) into functions. Hopefully this is a step towards making the test GUI code a bit less messy.


mkiael (approved):
Nice!

PR #47, opened 2021-10-02
css: Add support for expanding font-variant

mkiael:


css/parser.h
```diff
@@ -299,6 +302,13 @@ class Parser final : util::BaseParser {
         return std::nullopt;
     }
 
+    std::optional<std::string_view> try_parse_font_variant(Tokenizer &tokenizer) const {
+        if (auto maybe_font_variant = tokenizer.get(); *maybe_font_variant == "small-caps") {
```
robinlinden, line 306: Not sure if this can happen, but if we get here, and `tokenizer.get()` returns `nullopt`, this is UB. `maybe_font_variant.value()` is a throwing alternative. All the other if-statements in this seem to first check that `tokenizer.get()` doesn't return `nullopt` before dereferencing it.

robinlinden (approved):

css/parser.h
mkiael: Ah, this is not right! Good catch! Will fix.

PR #48, opened 2021-10-03
gfx: Port scrolling from an opengl translation to translating the painted rects

robinlinden:
This is probably slightly less efficient, but it means we will be able to cull all objects outside of our current viewport if we want.


mkiael:
Sounds and looks good!

mkiael (approved):

PR #64, opened 2021-10-15
css: Add support for parsing CSS comments

### LINKED ISSUE #37: Support comments when parsing CSS
https://developer.mozilla.org/en-US/docs/Web/CSS/Comments

robinlinden:
Resolves #37

### DIFF
--- a/css/parser.h
+++ b/css/parser.h
@@ -28,26 +28,18 @@ public:
         std::vector<css::Rule> rules;
         std::string_view media_query;
 
-        skip_whitespace();
+        skip_whitespace_and_comments();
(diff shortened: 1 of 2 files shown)


robinlinden:
Okay, so I didn't add support for comments everywhere they're allowed, but now CSS comments work everywhere we used skip_whitespace to support arbitrary whitespace. Maybe this is good enough for now? Most of the parser (including my comment-dodging code) will go away once we write a spec-compliant parser, so I don't think it's worth it to spend more time on it to try to be 100% compliant for comments right now.

PR #77, opened 2021-10-28
Refactor http get with socket wrapper

mkiael:


mkiael:
Great review. Think I got it all fixed now.

robinlinden (approved):

PR #80, opened 2021-10-31
Pimpl socket classes

mkiael:
This commit adds usage of the pimpl idiom in the socket classes. This prevents the asio dependency to leak into the other libraries.


robinlinden (approved):
Looks good! Stopping dependencies from leaking everywhere is always great. :P

PR #81, opened 2021-11-02
Receive http response in steps

mkiael:
Some refactoring to  parse a http response in steps. The reason for this is to be able to examine the headers before starting to receive and parse the body.

### DIFF
--- a/net/socket.cpp
+++ b/net/socket.cpp
@@ -32,6 +32,11 @@ std::string read_all_impl(auto &socket) {
     return data;
 }
 
+std::size_t read_until_impl(auto &socket, std::string &data, std::string_view delimiter) {
+    asio::error_code ec;
+    return asio::read_until(socket, asio::dynamic_buffer(data), delimiter, ec);
+}
+
 } // namespace
 
 struct Socket::Impl {
@@ -60,6 +65,10 @@ std::size_t Socket::write(std::string_view data) {
 
 std::string Socket::read_all() {
     return read_all_impl(impl_->socket);
+}
+
+std::size_t Socket::read_until(std::string &data, std::string_view delimiter) {
+    return read_until_impl(impl_->socket, data, delimiter);
 }
 
 struct SecureSocket::Impl {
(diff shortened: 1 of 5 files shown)


net/socket.cpp
```diff
@@ -32,6 +32,11 @@ std::string read_all_impl(auto &socket) {
     return data;
 }
 
+std::size_t read_until_impl(auto &socket, std::string &data, std::string_view delimiter) {
```
robinlinden, line 35: May be a nicer API to also return a std::string in this one? It looks like the way it's used in the commit after, every call to read_until is followed by clearing the data with erase and reading into the new empty string? We still get the size by calling string::size() when we need it.

protocol/http.cpp
```diff
@@ -92,28 +59,40 @@ std::string Http::create_get_request(uri::Uri const &uri) {
...
 }
 
-Response Http::parse_response(std::string_view data) {
-    auto [header, body] = split(data, "\r\n\r\n");
-    auto status_line = parse_status_line(header);
-    if (!status_line) {
-        return {Error::InvalidResponse};
+std::optional<StatusLine> Http::parse_status_line(std::string_view status_line) {
+    auto sep1 = status_line.find(' ');
+    if (sep1 == std::string_view::npos) {
```
robinlinden, line 64: Not needed, but you could also access `npos` from the `status_line` variable. That way you only need to change the argument type in one place if it ever changes. ...the argument type of this will never change, but it's a neat thing to be able to do. :P

robinlinden (approved):
Look good. Great refactoring!

net/socket.cpp
mkiael: I agree that this API is not great. The thing is that asio::read_until appends to the buffer (the string in this case) in chunks and returns once the delimiter is found. So the string may contain more data after the delimiter. The return value is not the number of bytes in the string, it is the position of the delimiter in the buffer (delimiter included). So the erase is only to remove the "consumed" data.

I got an idea while writing this on how we could do to make this return a string instead. I'll do a commit and add it to this PR.

protocol/http.cpp
mkiael: Fixed!

net/socket.cpp
robinlinden: Nice! Looks good! I should've looked at the asio API before commenting, but glad you found a solution. :P

robinlinden (approved):

PR #85, opened 2021-11-06
Add header for string utility functions

mkiael:
Moved function for case insensitive string comparison to the util library since I'll need it when comparing header names.
Suggestions for where to place this are welcome!

### DIFF
--- a/html2/tokenizer.cpp
+++ b/html2/tokenizer.cpp
@@ -4,9 +4,10 @@
 
 #include "html2/tokenizer.h"
 
+#include "util/string.h"
+
 #include <spdlog/spdlog.h>
 
-#include <cctype>
 #include <cstring>
 #include <exception>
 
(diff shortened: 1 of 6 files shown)


robinlinden (approved):
While expanding the `util` library is a bit sad, I think it's the best place we have for this right now.. I was looking adding a `core` library when implementing the `Generator` coroutine stuff, but that's really just a util library with another name, so I ended up creating a `coro` library for my thing. :x 

Let's let the string helpers live in util for now and we can rethink it later. What I did was just copy the code from `html` to `html2` since `html` is going away anyway, but if you're going to be using it in more libraries, sharing it and testing it is definitely better!

Can you switch to using this new function in https://github.com/robinlinden/hastur/blob/master/html/parser.h#L143-L154 too now that you're breaking it out into a library anyway?

html2/tokenizer.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "html2/tokenizer.h"
 
+#include "util/string.h"
+
 #include <spdlog/spdlog.h>
 
 #include <cctype>
```
robinlinden, line 11: I'm pretty sure `std::tolower` was the only thing from cctype, so this header can be nuked now!

mkiael: Fixed!

PR #87, opened 2021-11-07
Add function to trim whitespaces from start/end of string

mkiael:


util/string_test.cpp
```diff
@@ -25,5 +29,30 @@ int main() {
...
+        expect_eq(trim_start("word "), "word ");
+        expect_eq(trim_start("\r\n"), "");
+    });
+
+    etest::test("trim end", [] {
+        expect_eq(trim_end("abc "), "abc");
+        expect_eq(trim_end("53 \r\n"), "53");
+        expect_eq(trim_end("hello world!\t"), "hello world!");
+        expect_eq(trim_end(" word"), " word");
+        expect_eq(trim_start("\r\n"), "");
```
robinlinden, line 45: Should be calling `trim_end` and not `trim_start`?

util/string_test.cpp
```diff
@@ -25,5 +29,30 @@ int main() {
...
+    });
+
+    etest::test("trim", [] {
+        expect_eq(trim("abc"), "abc");
+        expect_eq(trim_start("\t431"), "431");
```
robinlinden, line 50: Should be calling `trim` and not `trim_start`?

util/string_test.cpp
```diff
@@ -25,5 +29,30 @@ int main() {
...
+        expect_eq(trim("53 \r\n"), "53");
+        expect_eq(trim("\t\thello world\n"), "hello world");
+        expect_eq(trim(" a b c d "), "a b c d");
+        expect_eq(trim_start("\r\n"), "");
```
robinlinden, line 54: Should be calling `trim` and not `trim_start`?

util/string.h
```diff
@@ -24,6 +27,22 @@ constexpr bool no_case_compare(std::string_view a, std::string_view b) {
     return true;
 }
 
+inline std::string trim_start(std::string s) {
+    s.erase(begin(s), std::find_if(begin(s), end(s), [](auto ch) { return !std::isspace(ch); }));
```
robinlinden, line 31: You can replace this with https://en.cppreference.com/w/cpp/string/basic_string/erase2 and not have to deal with iterators at all. Awesome new thing in C++20!

util/string.h
```diff
@@ -24,6 +27,22 @@ constexpr bool no_case_compare(std::string_view a, std::string_view b) {
...
+    return s;
+}
+
+inline std::string trim_end(std::string s) {
+    s.erase(std::find_if(rbegin(s), rend(s), [](auto ch) { return !std::isspace(ch); }).base(), end(s));
+    return s;
+}
+
+inline std::string trim(std::string s) {
+    s = trim_start(s);
```
robinlinden, line 41: You could move s into these.

robinlinden (changes requested):

util/string_test.cpp
mkiael: That was sloppy! Good catch!

util/string_test.cpp
mkiael: Good catch!

util/string.h
mkiael: Oh, nice. I'll fix!

robinlinden (approved):
Looks good!

PR #89, opened 2021-11-10
ci: Update clang-12 to run everything under Valgrind

robinlinden:
The clang-12 job was chosen by a fair roll of the dice. :P
 Suggestions on what flags or whatever to pass to valgrind are very welcome!


mkiael (approved):
Great!

PR #92, opened 2021-11-13
Simplify no case compare

mkiael:
This PR adds the range-v3 library as a dependency. Unfortunately a patch is needed to be able to build the latest release of the library with clang13 and libc++. The patch is a commit from the master branch of the main repo.

Please object if you don't want that we add this dependency or if you have any concern with how the patching is handled.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -84,6 +84,13 @@ http_archive(
 )
 
 http_archive(
+    name = "range-v3",
+    sha256 = "3575e4645cd1a7d42fa42a6b016e75a7c72d31d13f72ee4e5bb9773d36303258",
+    strip_prefix = "range-v3-83783f578e0e6666d68a3bf17b0038a80e62530e",
+    url = "https://github.com/ericniebler/range-v3/archive/83783f578e0e6666d68a3bf17b0038a80e62530e.tar.gz",
+)
+
+http_archive(
     name = "sfml",  # Zlib
     build_file = "//third_party:sfml.BUILD",
     sha256 = "6124b5fe3d96e7f681f587e2d5b456cd0ec460393dfe46691f1933d6bde0640b",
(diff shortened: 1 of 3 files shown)


util/BUILD
```diff
@@ -7,6 +7,7 @@ cc_library(
         "string.h",
     ],
     visibility = ["//visibility:public"],
+    deps = ["@range-v3//:range-v3"],
```
robinlinden, line 10: Is it possible to use `@ranges-v3//:std` to ensure that we can switch to the stdlib ranges once they're available in all compilers/stdlib combos we support?

WORKSPACE
```diff
@@ -122,3 +122,12 @@ http_archive(
     strip_prefix = "zlib-1.2.11",
     url = "https://github.com/madler/zlib/archive/v1.2.11.tar.gz",
 )
+
+http_archive(
+    name = "range-v3",
+    patch_args = ["-p1"],
+    patches = ["//third_party/patches:range-v3.patch"],
```
robinlinden, line 129: Looking at https://github.com/ericniebler/range-v3/compare/0.11.0...master I think we should just go with the latest commit on master. :P Everything after 0.11.0 appears to be bugfixes.

You can see an example of including an archive of a specifc commit if you look at stb in this file.

robinlinden: I also think I have the libraries in here in alphabetical order (within their group.)

robinlinden (changes requested):

util/BUILD
robinlinden, line 10: This could be simplified to `@ranges-v3`. If the package name matches the dependency's name, it's not needed!

robinlinden (approved):
Looks great! Looking forward to start using ranges everywhere. :D

robinlinden (commented):
Oh, actually the commit message from the deps commit is wrong. It's still talking patching the dependency. :x

mkiael:
Fixed the commit message, good catch!

PR #93, opened 2021-11-14
util: Fix UB in trim

robinlinden:
std::isspace accepts an int, but it's UB to pass anything that's not EOF
or representable as an unsigned char.

See: https://en.cppreference.com/w/cpp/string/byte/isspace

Without the fix, the test added fails in a MSVC debug build (due to an assertion checking that the int passed is within the acceptable range), but appears to pass in all CI jobs even without the fix due to none of them being debug builds.

Here is an example of a failed CI run (with --config debug added): https://github.com/robinlinden/hastur/runs/4204736206?check_suite_focus=true

### DIFF
--- a/util/string.h
+++ b/util/string.h
@@ -20,12 +20,12 @@ constexpr bool no_case_compare(std::string_view a, std::string_view b) {
 }
 
 inline std::string trim_start(std::string s) {
-    s.erase(begin(s), std::find_if(begin(s), end(s), [](auto ch) { return !std::isspace(ch); }));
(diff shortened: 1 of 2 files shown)


mkiael:
Good work finding this!

mkiael (approved):

robinlinden:
I did the same thing myself because I didn't know this was how std::isspace worked: https://github.com/robinlinden/hastur/commit/d0f197ff5c3da96d1b70f84b0fe95a32aad02d04
Very surprising behaviour. :D

PR #95, opened 2021-11-15
Add class for handling headers

mkiael:
Adds a class for handling headers from a http response. Value look up is case insensitive.


robinlinden (approved):
Looks good!

PR #97, opened 2021-11-18
Fix for returning wrong value from FakeSocket::read_bytes

mkiael:


robinlinden (approved):
Whoops. I should have caught that. :( I was putting too much trust in our unused-variable warnings catching this kind of thing, but I guess those warnings only trigger on simple types with no side-effects.

PR #102, opened 2021-11-25
Follow links when clicked

robinlinden:
This isn't super useful without a few features still missing, but it makes my testing different pages a lot more convenient.


mkiael (approved):
Looks good!

PR #104, opened 2021-11-27
Start implementing spec-compliant tree construction

robinlinden:
This isn't enough to fully handle http://example.com yet, but it's enough to get us into the `<title>` element, and the diff was getting a bit big.

### DIFF
--- a/html2/tree_builder.cpp
+++ b/html2/tree_builder.cpp
@@ -1,0 +1,231 @@
+// SPDX-FileCopyrightText: 2021 Robin Lindén <dev@robinlinden.eu>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "html2/tree_builder.h"
+
+#include "dom2/document_type.h"
+
+#include <spdlog/spdlog.h>
+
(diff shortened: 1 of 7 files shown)


mkiael:
Missing copyright header in `html2/tree_builder.h` and `html2/tree_builder.cpp`

html2/tree_builder.cpp
```diff
@@ -0,0 +1,229 @@
+#include "html2/tree_builder.h"
+
+#include "dom2/document_type.h"
+
+#include <spdlog/spdlog.h>
+
+#include <exception>
+#include <functional>
+
+using namespace std::literals;
```
mkiael, line 10: Did not see anything that was used from this namespace. But I might have missed something :)

mkiael (approved):

html2/tree_builder.cpp
robinlinden: No, you're right. :D I probably did use it at some point, but I removed a lot of wip things when cleaning this before pushing it. Good catch!

PR #105, opened 2021-11-28
gfx+render: Move viewport setup from render to gfx

robinlinden:
It felt a bit weird that render had a function that you needed to call
to get the OpenGL painter to work correctly. This still doesn't feel
perfect, but at least now all OpenGL functionality is gathered in one
place.

After this change, gfx only deals with the graphics abstraction and
render only deals with calling into gfx based on what layout nodes it
encounters.


mkiael (approved):
Nice!

PR #109, opened 2021-11-30
dom2: Implement CharacterData and friends

robinlinden:
This implements the CharacterData methods as well as fill out the
simple classes that directly inherit from CharacterData.

I'm not sure what the deal with ProcessingInstruction is. The spec
specifies a member and a getter, but no setter or constructor, so we'll
just have to see how it's used in the rest of the spec when we get there
I suppose.

### DIFF
--- a/dom2/character_data_test.cpp
+++ b/dom2/character_data_test.cpp
@@ -1,0 +1,88 @@
+// SPDX-FileCopyrightText: 2021 Robin Lindén <dev@robinlinden.eu>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "dom2/character_data.h"
+
+#include "etest/etest.h"
+
+#include <cstddef>
+#include <string>
+#include <string_view>
+#include <utility>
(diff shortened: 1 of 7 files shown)


dom2/character_data_test.cpp
```diff
@@ -0,0 +1,86 @@
+// SPDX-FileCopyrightText: 2021 Robin Lindén <dev@robinlinden.eu>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "dom2/character_data.h"
+
+#include "etest/etest.h"
+
+#include <string>
+#include <string_view>
```
mkiael, line 10: Should `<cstddef>` and `<utility>` be added?

mkiael (approved):

dom2/character_data_test.cpp
robinlinden: You're right, they should be! :x `std::size_t` and `std::move`.

PR #110, opened 2021-12-02
protocol: Add initial support for decoding a chunked body

mkiael:
Initial support for decoding a chunked body. Still missing support for parsing chunk extensions and trailer part.

Part of #28

### DIFF
--- a/protocol/http.h
+++ b/protocol/http.h
@@ -7,12 +7,15 @@
 #define PROTOCOL_HTTP_H_
 
 #include "uri/uri.h"
+#include "util/string.h"
 
+#include <charconv>
 #include <cstddef>
 #include <map>
 #include <optional>
 #include <string>
 #include <string_view>
+#include <system_error>
 #include <utility>
 
 namespace protocol {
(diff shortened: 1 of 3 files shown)


protocol/http.h
```diff
@@ -54,34 +57,90 @@ struct Response {
...
             if (data.empty()) {
                 return {Error::InvalidResponse};
             }
             auto headers = Http::parse_headers(data.substr(0, data.size() - 4));
             if (headers.size() == 0) {
                 return {Error::InvalidResponse};
             }
-            data = socket.read_all();
+            auto encoding = headers.get("transfer-encoding"sv);
+            if (encoding.has_value() && *encoding == "chunked"sv) {
```
robinlinden, line 82: This can be written as `if (encoding != "chunked"sv) {`. `std::optional`'s comparison operators internally check if it has a value, and if not, returns false when compared to something with a value.

protocol/http_test.cpp
```diff
@@ -153,6 +166,102 @@ int main() {
...
+        auto socket = create_chunked_socket(
+                "5\r\nhello"
+                "0\r\n\r\n");
+
+        auto response = protocol::Http::get(socket, create_uri());
+
+        expect_eq(response.err, protocol::Error::InvalidResponse);
+    });
+
+    etest::test("transfer-encoding chunked, chunk to short", [] {
```
robinlinden, line 245: `to` -> `too`, same with the next test.

robinlinden (approved):
Looks great!

PR #111, opened 2021-12-04
Bless dom with a saner node structure

robinlinden:
This changes it so that Text nodes don't have an always-empty
children member, which means that Node can now be a alias for
std::variant<Element, Text> instead of a type of its own.

I ran into a lot of issues with dom::Node being implicitly constructed
from dom::Element, leading to us returning pointers or references to
temporaries, so the functions I ran into issues with were converted to
use std::reference_wrapper instead of just a plain reference. This means
that it's a compile-time error to pass a temporary.


mkiael (approved):
Nice!

PR #113, opened 2021-12-06
Add split utility functions

mkiael:


util/string.h
```diff
@@ -12,13 +12,32 @@
...
+#include <utility>
+#include <vector>
 
 namespace util {
 
 constexpr bool no_case_compare(std::string_view a, std::string_view b) {
     return ranges::equal(a, b, [](auto c1, auto c2) { return std::tolower(c1) == std::tolower(c2); });
 }
 
+inline std::vector<std::string_view> split(std::string_view str, std::string_view sep) {
```
robinlinden, line 24: I **think** these can be constexpr!

util/string_test.cpp
```diff
@@ -10,7 +10,10 @@ using namespace std::literals;
 
 using etest::expect;
 using etest::expect_eq;
+using etest::require;
 using util::no_case_compare;
+using util::split;
```
robinlinden, line 15: I have nothing against `using namespace util;` to keep this list short. Do you think listing each function individually is better for readability? If we had one test-suite per function I think it'd be clearer, but when we have one larger test-suite for everything in a lib, I'm not so sure.

robinlinden (approved):
Looks great!

util/string.h
mkiael: Seems like constexpr std::vector is only supported by MSVC at the moment. There is support in trunk builds of GCC and Clang, so I guess we'll have to wait :)

util/string_test.cpp
mkiael: I agree, I'll add `using namespace util;`.

PR #115, opened 2021-12-08
Use string_view in trim functions

mkiael:
This commit changes to use string_view in all trim functions. Also made these function constexpr.

### DIFF
--- a/util/string.h
+++ b/util/string.h
@@ -38,6 +38,11 @@ inline std::pair<std::string_view, std::string_view> split_once(std::string_view
     return {str, ""};
 }
 
+constexpr bool is_whitespace(char ch) {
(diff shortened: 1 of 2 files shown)


mkiael:
Forgot to add tests for `is_whitespace()`. Will fix.

robinlinden (approved):
Looks good! I think you can get rid of the `<string>` and `<utility>` includes from `util/string.h` now as well?

robinlinden (approved):
Rebase instead of merging when updating the PR against master (I think I asked GitHub to require a linear history, so hopefully it can't even be merged with merge commits?), and I think it's nicer to add the unit test in the same commit as you added the feature you're testing.

mkiael:
Sorry, got a bit messy. Fixed now!

PR #117, opened 2021-12-11
ci: Save the Bazel cache across runs

robinlinden:
Seems like the GH Actions cache is down everywhere right: `https://github.com/actions/cache/issues/698`, but I think should work. Not 100% sure about the Windows cache since the output location is a bit weird, but we'll find out once the cache service is available again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,6 +68,11 @@ jobs:
       run: |
           sudo apt update
           sudo apt install libxrandr-dev libgl-dev ${{ matrix.compiler }}-${{ matrix.version }} ${{ matrix.apt }}
+    - uses: actions/cache@v2
+      with:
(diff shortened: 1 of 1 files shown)


mkiael (approved):

mkiael (approved):

robinlinden:
This has some issues. The MSVC cache location is incorrect, and the created cache is never updated. Will fix in a bit!

robinlinden:
I dropped MSVC from this PR and updated the Linux CI cache so that it's refreshed whenever we touch `WORKSPACE` or anything in `third_party`. This is hopefully good enough and will at least ensure that our third-party code and downloads will always be in the cache.

We should definitely have another look at setting up a cache for the MSVC job again though. After this change, that job takes ~5 minutes longer than any other job.

mkiael (approved):

PR #119, opened 2021-12-15
gfx: Add a simple example application

robinlinden:
@mkiael Do you think doing something like this is alright? It's pretty helpful when expanding the gfx API with e.g. text rendering. :P


mkiael:
This is a good idea, makes it easy to experiment.

mkiael (approved):

PR #121, opened 2021-12-20
Split up `//util`

robinlinden:
I didn't like that when using the string utils, you also got very unrelated parsing utils, so this commit splits the monolith util target into 1 target per util.

I also played around a bit with Starlark magic in the build file. Not sure if it simplified it or not, but PTAL! It'll need a minor change if we add utils with .cpp-files, but this works as things are right now.

Now instead of depending on `//util`, you depend on `//util:string` or `//util:base_parser`.

### DIFF
--- a/css/BUILD
+++ b/css/BUILD
@@ -15,7 +15,7 @@ cc_library(
     data = ["default.css"],
     visibility = ["//visibility:public"],
     deps = [
-        "//util",
+        "//util:base_parser",
         "@fmt",
     ],
 )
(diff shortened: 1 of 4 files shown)


mkiael (approved):
I think splitting this up is a good change!

The changes in the build file definitely makes it simpler to add a new util library, but maybe there is a slight increase in cognitive load when reading it. But that's fine by me! Interesting to see how Starlark can be used in this way.

PR #123, opened 2021-12-21
html2: Add support for parsing comments

mkiael:
I think we need to look into how the eof check is done to get the eof tokens emitted properly. Left two tests commented out until this has been fixed.

### DIFF
--- a/html2/tokenizer_test.cpp
+++ b/html2/tokenizer_test.cpp
@@ -14,28 +14,128 @@
 #include <vector>
 
 using namespace std::literals;
+
 using etest::expect_eq;
 using etest::require;
+
+using namespace html2;
+
+namespace {
+std::vector<Token> run_tokenizer(std::string_view input) {
+    std::vector<Token> tokens;
+    Tokenizer tokenizer{input, [&](Token &&t) {
+                            tokens.push_back(std::move(t));
+                        }};
+    tokenizer.run();
+    return tokens;
+}
+} // namespace
 
 int main() {
     etest::test("simple_page", [] {
         std::ifstream page{"html2/test/simple_page.html", std::ios::binary};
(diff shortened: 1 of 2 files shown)


html2/tokenizer_test.cpp
```diff
@@ -14,28 +14,128 @@
...
 
 using namespace std::literals;
+
 using etest::expect_eq;
 using etest::require;
 
+using namespace html2;
+
+namespace {
+std::vector<Token> run_tokenizer(std::string_view input) {
```
robinlinden, line 24: Nice!

html2/tokenizer_test.cpp
```diff
@@ -14,28 +14,128 @@
...
+        expect_eq(tokens, std::vector<Token>{CommentToken{.data = ""}});
+    });
+
+    etest::test("comment, incorrectly closed comment", [] {
+        auto tokens = run_tokenizer("<!--abc--!>");
+
+        expect_eq(tokens, std::vector<Token>{CommentToken{.data = "abc"}});
+    });
+
+    // TODO(mkiael): Enable when eof check is fixed
```
robinlinden, line 126: What needs to be done with the eof check?

robinlinden (approved):
Looks great!

html2/tokenizer_test.cpp
mkiael: I'm not sure, but it might be that the `!is_eof()` in the while condition makes the loop break to early, causing us to miss emitting eof tokens.

robinlinden: Ah, yeah, you're probably right. Maybe we should be setting a flag if we encounter an eof-token in emit or something. :P

PR #125, opened 2021-12-26
Support (integer) scaling the entire application

robinlinden:
I was having issues working on hastur on hidpi displays, and this is a decent step in the direction of proper support for anything but peasant resolutions. :P


mkiael (approved):
Great!

PR #126, opened 2021-12-28
Fix so that eof token is emitted from tokenizer

mkiael:
Not sure if `while (true)` is a good idea. My gut feeling tells me that a complete tokenizer never should end up in an eternal loop. Opening PR for discussion.


robinlinden (approved):

robinlinden:
Yeah, I don't think the tokenizer should ever enter a loop once we're done with it. For now we can attach a debugger and check where it's looping if needed. :P

PR #129, opened 2021-12-31
Fix out-of-bounds read in the new HTML tokenizer

robinlinden:
Also set up the MSVC CI configuration so that it detects this kind of issue.

Example of it detecting the issue fixed in this PR here: https://github.com/robinlinden/hastur/actions/runs/1641134299

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,11 +104,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build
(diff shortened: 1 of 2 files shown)


mkiael:
Ah, may bad. I should have catch this!

mkiael (approved):

robinlinden:
No worries! I'm surprised literally none of the CI caught it. :P I mean, we run with both valgrind and asan. Maybe we need some `_FORTIFY_SOURCE`-like thing.

PR #130, opened 2022-01-01
Add support for tokenizing HTML entities

robinlinden:
~Based on #129 since the tokenizer doesn't really work without it.~ Rebased on master since CI passes even with the out-of-bounds reads.


html2/tokenizer.cpp
```diff
@@ -1,15 +1,19 @@
...
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include "html2/tokenizer.h"
 
+#include "html2/character_reference.h"
 #include "util/string.h"
 
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
```
mkiael, line 12: Is this include needed?

mkiael (approved):

html2/tokenizer.cpp
robinlinden: Nope! Forgot to remove it when simplifying my code. Good catch, removed!

html2/tokenizer.cpp
```diff
@@ -804,4 +892,20 @@ void Tokenizer::reconsume_in(State state) {
...
+
+void Tokenizer::flush_code_points_consumed_as_a_character_reference() {
+    if (consumed_as_part_of_an_attribute()) {
+        current_attribute().value += temporary_buffer_;
+        return;
+    }
+
+    for (char c : temporary_buffer_) {
+        emit(CharacterToken{c});
+    }
```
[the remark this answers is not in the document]
robinlinden: Ah, yeah, didn't notice that. Broken out into a separate function. Good idea!

PR #134, opened 2022-01-06
Add support for tokenizing non-ascii html entities

robinlinden:
Maybe writing a unicode code point -> utf8 decoder is a bit silly, but it was fun and it has a tiny API, so we can easily replace it with a library in the future if we feel like it.


mkiael (approved):
Very nice!

PR #136, opened 2022-01-08
Set tokenizer state

mkiael:
This PR adds possibility to set the state of the tokenizer. The tree builder will need to be able to set this during processing of tokens. Will be covered by tests when ongoing work on script tag is done.

### DIFF
--- a/html2/tokenizer.h
+++ b/html2/tokenizer.h
@@ -147,7 +147,7 @@ std::string to_string(Token const &);
 
 class Tokenizer {
 public:
-    Tokenizer(std::string_view input, std::function<void(Token &&)> on_emit)
+    Tokenizer(std::string_view input, std::function<void(Token &&, Tokenizer &)> on_emit)
         : input_{input}, on_emit_{std::move(on_emit)} {}
 
     void set_state(State);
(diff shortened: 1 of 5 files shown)


html2/tokenizer.h
```diff
@@ -147,8 +147,10 @@ std::string to_string(Token const &);
 
 class Tokenizer {
 public:
-    Tokenizer(std::string_view input, std::function<void(Token &&)> on_emit)
+    Tokenizer(std::string_view input, std::function<void(Token &&, Tokenizer &)> on_emit)
         : input_{input}, on_emit_{std::move(on_emit)} {}
+
+    void set_state(State state);
```
robinlinden, line 153: If you want to, you can drop `state` from this since it's obvious from the signature what it is anyway.

robinlinden (approved):

html2/tokenizer.h
mkiael: True! Fixed

robinlinden: But not pushed? :O

mkiael: Ah, I assumed you meant the name of the function.

mkiael: Now it's fixed for real.

robinlinden: Ah, sorry, I edited the comment to specify the name of the argument, but I was too slow. :P I think the name of the function is perfectly fine as `set_state` since that's what you expect a setter to be called.

PR #137, opened 2022-01-11
build: Enable Bazel's layering check

robinlinden:
This checks that our targets aren't relying on transitive dependencies.


mkiael (approved):
Wow, this is very cool!

PR #138, opened 2022-01-11
html2: Add support for parsing script tag

mkiael:
This PR implements all states related to the script tag. Still a todo for emitting replacement character for null.
`TreeBuilder` needs to be updated to set the state of the `Tokenizer` when a `<script>` start tag is encountered.

Part of #139

### DIFF
--- a/html2/tokenizer.cpp
+++ b/html2/tokenizer.cpp
@@ -122,6 +122,28 @@ void Tokenizer::run() {
                 break;
             }
 
+            case State::ScriptData: {
+                auto c = consume_next_input_character();
+                if (!c) {
+                    emit(EndOfFileToken{});
+                    return;
+                }
+
+                switch (*c) {
+                    case '<':
+                        state_ = State::ScriptDataLessThanSign;
+                        continue;
+                    case '\0':
+                        // This is an unexpected-null-character parse error.
+                        // TODO(mkiael): Emit replacement character for null
+                        // emit(CharacterToken("\xFF\xFD"));
+                        continue;
+                    default:
+                        emit(CharacterToken{*c});
+                        continue;
+                }
+            }
+
             case State::TagOpen: {
                 auto c = consume_next_input_character();
                 if (!c) {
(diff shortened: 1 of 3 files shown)


html2/tokenizer.cpp
```diff
@@ -912,4 +1439,11 @@ void Tokenizer::emit_temporary_buffer_as_character_tokens() {
     }
 }
 
+bool Tokenizer::is_appropriate_end_tag_token(Token token) const {
```
robinlinden, line 1442: `Token const &`

html2/tokenizer.cpp
```diff
@@ -214,6 +236,507 @@ void Tokenizer::run() {
...
+
+                if (!c) {
+                    anything_else();
+                    continue;
+                }
+
+                if (is_ascii_upper_alpha(*c)) {
+                    auto lower_case_c = to_lower(*c);
+                    std::get<EndTagToken>(current_token_).tag_name.append(1, lower_case_c);
+                    temporary_buffer_.append(1, lower_case_c);
```
robinlinden, line 296: "Append the current input character to the temporary buffer.", so this probably shouldn't be lower_case_c?
https://html.spec.whatwg.org/multipage/parsing.html#script-data-end-tag-name-state

html2/tokenizer.cpp
```diff
@@ -214,6 +236,507 @@ void Tokenizer::run() {
...
+                }
+
+                switch (*c) {
+                    case '\t':
+                    case '\n':
+                    case '\f':
+                    case ' ':
+                    case '/':
+                    case '>':
+                        if (temporary_buffer_ == "script"s) {
```
robinlinden, line 583: When comparing you can compare to a `""sv` instead. No need to require an heap-allocated string for this.

html2/tokenizer.cpp
```diff
@@ -214,6 +236,507 @@ void Tokenizer::run() {
...
+
+            case State::ScriptDataDoubleEscapeEnd: {
+                auto c = consume_next_input_character();
+                if (!c) {
+                    reconsume_in(State::ScriptDataDoubleEscaped);
+                    continue;
+                }
+
+                if (is_ascii_upper_alpha(*c)) {
+                    auto lower_case_c = to_lower(*c);
```
robinlinden, line 708: I think it's more readable if you inline the `to_lower` into the `append` call. No need for a temporary variable when it's only used once.

html2/tokenizer.cpp
```diff
@@ -214,6 +236,507 @@ void Tokenizer::run() {
...
+                }
+
+                switch (*c) {
+                    case '\t':
+                    case '\n':
+                    case '\f':
+                    case ' ':
+                    case '/':
+                    case '>':
+                        if (temporary_buffer_ == "script"s) {
```
robinlinden, line 727: `""sv` instead of `""s`

robinlinden (changes requested):
Looks good in general with some minor comments!

It would also be nice (but not required!) if you could add another test to get the codecov patch check to pass since there are a few implemented but not tested states as well. Would probably only take 1 or 2 tests or something?
`codecov/patch Failing after 1s — 72.92% of diff hit (target 74.92%)`

html2/tokenizer.cpp
```diff
@@ -246,6 +769,7 @@ void Tokenizer::run() {
                     continue;
                 }
 
+                // Should append lower case!
```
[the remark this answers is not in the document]
mkiael: Yes, fixed!

html2/tokenizer.cpp
mkiael: Good catch! Fixed!

html2/tokenizer.cpp
```diff
@@ -214,6 +236,507 @@ void Tokenizer::run() {
...
+
+                if (!c) {
+                    anything_else();
+                    continue;
+                }
+
+                if (is_ascii_upper_alpha(*c)) {
+                    auto lower_case_c = to_lower(*c);
+                    std::get<EndTagToken>(current_token_).tag_name.append(1, lower_case_c);
+                    temporary_buffer_.append(1, lower_case_c);
```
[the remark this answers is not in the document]
mkiael: Good catch! Fixed!

html2/tokenizer.cpp
robinlinden: Did you intend to change this? :P

robinlinden (approved):
Looks good! Only a nit left. Up to you if you change it or not!

html2/tokenizer.cpp
mkiael: I was! Fixed now.

PR #140, opened 2022-01-16
Fix parsing of uppercase attribute name

mkiael:
This PR fixes so that the lowercase version of the input character is appended to the attribute name. Also adds `continue` to break the loop after the character has been appended.

### DIFF
--- a/html2/tokenizer_test.cpp
+++ b/html2/tokenizer_test.cpp
@@ -458,5 +458,34 @@ int main() {
         expect_eq(glyph, "∾̳"sv);
     });
 
+    etest::test("attribute, one attribute single quoted", [] {
+        auto tokens = run_tokenizer("<tag a='b'>");
+
+        expect_token(tokens, StartTagToken{.tag_name = "tag", .attributes = {{"a", "b"}}});
+        expect_token(tokens, EndOfFileToken{});
+    });
+
+    etest::test("attribute, one attribute double quoted", [] {
+        auto tokens = run_tokenizer("<tag a=\"b\">");
+
+        expect_token(tokens, StartTagToken{.tag_name = "tag", .attributes = {{"a", "b"}}});
+        expect_token(tokens, EndOfFileToken{});
+    });
+
+    etest::test("attribute, one uppercase attribute", [] {
+        auto tokens = run_tokenizer("<tag ATTRIB=\"ABC123\">");
+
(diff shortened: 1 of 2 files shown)


html2/tokenizer_test.cpp
```diff
@@ -458,5 +458,34 @@ int main() {
         expect_eq(glyph, "∾̳"sv);
     });
 
+    etest::test("attribute, one attribute single qouted", [] {
+        auto tokens = run_tokenizer("<tag a=\'b\'>");
```
robinlinden, line 462: You don't have to escape `'` inside `"` strings.

html2/tokenizer_test.cpp
```diff
@@ -458,5 +458,34 @@ int main() {
...
+
+    etest::test("attribute, one uppercase attribute", [] {
+        auto tokens = run_tokenizer("<tag ATTRIB=\"ABC123\">");
+
+        expect_token(tokens, StartTagToken{.tag_name = "tag", .attributes = {{"attrib", "ABC123"}}});
+        expect_token(tokens, EndOfFileToken{});
+    });
+
+    etest::test("attribute, multiple attributes", [] {
+        auto tokens = run_tokenizer("<tag  foo=\"bar\" A=\'B\'  value=\'321\'>");
```
robinlinden, line 483: No need to escape `'`.

html2/tokenizer_test.cpp
```diff
@@ -458,5 +458,34 @@ int main() {
         expect_eq(glyph, "∾̳"sv);
     });
 
+    etest::test("attribute, one attribute single qouted", [] {
```
robinlinden, line 461: `qouted` -> `quoted`, same goes for the next test.

robinlinden (approved):
Looks good aside from some minor typos and unnecessary escaping.
Great job finding that bug!

html2/tokenizer_test.cpp
mkiael: Fixed!

html2/tokenizer_test.cpp
mkiael: Fixed!

html2/tokenizer_test.cpp
mkiael: Fixed!

PR #146, opened 2022-02-22
Add an SFML gfx backend

robinlinden:
This can currently only be tested by passing `--sf` to the gfx_example application.

The plan is to hack together a temporary text rendering in this to unblock some other parts we want to work on while I set up more proper text rendering.


mkiael (approved):
Nice!

PR #151, opened 2022-03-11
Add bits the text rendering depends on

robinlinden:
The most interesting change is that you can now hit F2 to switch between the SFML and GL rendering backend in the browser GUI.


mkiael (approved):
Awesome!

robinlinden:
I think this is the last of the small PRs before I actually PR text rendering. :P

PR #161, opened 2022-03-27
layout: Calculate em sizes based on the current node's font-size

robinlinden:
I need some bits for grabbing properties from the parent node for text rendering, and this is part of being able to test that in a nice way.


mkiael (approved):
Nice!

PR #168, opened 2022-04-24
html2: Don't drop tags closed immediately after an attribute name

robinlinden:


mkiael (approved):
Great finding this! :)

PR #184, opened 2022-05-15
html2: Set up infra for emitting tokenizer errors

robinlinden:
@mkiael What do you think of something like this for support doing something with tokenizer errors?


mkiael:
This looks great! Simple and clean as it should be :)

mkiael (approved):

PR #190, opened 2022-05-17
Make sure all tokens are consumed in html2 tests

robinlinden:
@mkiael Not sure if 005cb7827fc9d827cd15ad5f6a69ad1e57fc6be6 is a good idea. On one hand, I guess our tests would catch any behaviour change, but on the other hand https://en.cppreference.com/w/cpp/language/translation_phases#Phase_5 makes it sound like we shouldn't rely on literals being utf8 at runtime.

Maybe replace the text expected with the hex representation instead to be safe?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text=auto eol=lf
+* text=auto
--- a/html2/BUILD
+++ b/html2/BUILD
@@ -16,10 +16,6 @@ cc_library(
     ],
 )
 
-data_files = {
-    "tokenizer": ["test/simple_page.html"],
-}
-
 dependencies = {
     "tree_constructor": ["//dom2"],
 }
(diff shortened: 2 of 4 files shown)


mkiael:
> @mkiael Not sure if [005cb78](https://github.com/robinlinden/hastur/commit/005cb7827fc9d827cd15ad5f6a69ad1e57fc6be6) is a good idea. On one hand, I guess our tests would catch any behaviour change, but on the other hand https://en.cppreference.com/w/cpp/language/translation_phases#Phase_5 makes it sound like we shouldn't rely on literals being utf8 at runtime.
> 
> Maybe replace the text expected with the hex representation instead to be safe?

I vote on taking the safest choice here using a hex representation instead. Nice work cleaning up these tests!

robinlinden:
>Nice work cleaning up these tests!

Nice job building the abstraction that made it so clean!

robinlinden:
Right, updated the glyph construction commit to only compare against the UTF-8 hex representation! :P

mkiael (approved):

PR #192, opened 2022-05-19
gfx: Fix crash when font path does not exist

mkiael:
This commit makes so that any filesystem_error exception that is thrown from recursive_directory_iterator is catched.

### DIFF
--- a/gfx/sfml_painter.cpp
+++ b/gfx/sfml_painter.cpp
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2022 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2022 Mikael Larsson <c.mikael.larsson@gmail.com>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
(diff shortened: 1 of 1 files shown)


gfx/sfml_painter.cpp
```diff
@@ -20,10 +20,20 @@
 namespace gfx {
 namespace {
 
+std::filesystem::recursive_directory_iterator get_font_dir_iterator(std::filesystem::path const &path) {
```
robinlinden, line 24: Maybe use a function-try-block here and return directly? Makes it a bit more obvious what's going on imo.

```
auto get_font_dir_iterator(auto const &path) try {
  return iterator(path);
} catch (...) {
  return {};
}
```

robinlinden (approved):
Whoops! :P Guess I should've tested a bit more.

Maybe we should look into at least starting the browser or the gfx example in CI as a smoke test in the future? Should work with e.g. xvfb, and we could add a cmdline flag to exit after x seconds for testing.

robinlinden:
typo: `catched` -> `caught` in the commit message.

gfx/sfml_painter.cpp
mkiael: Yes, that looks nicer!

gfx/sfml_painter.cpp
```diff
@@ -20,10 +21,18 @@
 namespace gfx {
 namespace {
 
+auto get_font_dir_iterator(std::filesystem::path const &path) try {
+    return std::filesystem::recursive_directory_iterator(path);
+} catch (std::filesystem::filesystem_error const &) {
+    return std::filesystem::recursive_directory_iterator();
+} catch (...) {
+    throw;
+}
```
mkiael, line 30: Might be good to not silence all exceptions, only `filesystem_error`. Ok?

robinlinden: Yeah, probably best to only silence filesystem_error since we know it happens and that it's harmless when it does. No need to catch and rethrow everything else since we're not doing anything with it though!

mkiael: Ah, yea, of course!

robinlinden (approved):

Issue #193, opened 2022-05-19
Handle "auto" width


robinlinden:

robinlinden:
I think this is done. Will open a new issue if there's something more specific to do.

PR #199, opened 2022-05-22
layout: Update calculations for width and margin to handle auto values

mkiael:
Part of #193


robinlinden (approved):
Nice! Looks great! I see a lot of websites look at way better with this patch too!

PR #200, opened 2022-05-23
layout: Add calculation of border width

mkiael:


mkiael:
Just realized that `border-style` is a shorthand. Will change so that the style of every edge is checked.

layout/layout.cpp
```diff
@@ -199,6 +199,31 @@ void calculate_padding(LayoutBox &box, int const font_size) {
     }
 }
 
+void calculate_border(LayoutBox &box, int const font_size) {
+    std::string_view default_style = "none";
+    std::string_view default_width = "medium";
```
robinlinden, line 204: `to_px` doesn't handle named widths right now, so right now `thin` == `medium` == `thick` == `0`. Maybe default this to `3px` (the spec says that thin, medium, and thick all are implementation-defined) or something for now so things that want a border don't get 0?

robinlinden (approved):
Nice!

layout/layout.cpp
mkiael: Yes, that's a good idea. Will fix.

PR #206, opened 2022-05-26
html2: Add support for tokenizing the doctype public identifier

robinlinden:
Part of #207


robinlinden:
Once #208 is merged I can combine the double- and single-quoted tests, but I think this is ready for review now. :P

mkiael (approved):
Great work!

Issue #99, opened 2021-11-20
Handle borders


robinlinden:

robinlinden:
@mkiael Is this done enough that we can close it and open more specific issues if there's something we want to stick in the backlog?

robinlinden:
This is probably done. We can open more specific issues if there's anything missing.

PR #209, opened 2022-05-28
Render solid borders

mkiael:
This PR adds initial support for rendering solid borders. This might need to be refactored a bit once we start adding support for more border styles and border radius.

Part #99

### DIFF
--- a/render/render.cpp
+++ b/render/render.cpp
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2021-2022 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2022 Mikael Larsson <c.mikael.larsson@gmail.com>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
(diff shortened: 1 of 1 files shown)


render/render.cpp
```diff
@@ -104,6 +108,42 @@ void render_text(gfx::IPainter &painter, layout::LayoutBox const &layout, dom::T
...
 }
 
+void render_borders(gfx::IPainter &painter, layout::LayoutBox const &layout) {
+    // TODO(mkiael): Handle a lot more border styles
+    auto color = style::get_property(*layout.node, "color"sv).value_or(kDefaultColor);
+    auto border_box = layout.dimensions.border_box();
+    auto const &border_size = layout.dimensions.border;
+    if (border_size.left > 0) {
+        geom::Rect border{border_box.x,
+                border_box.y + border_size.top,
```
robinlinden, line 118: (0,0) is the top left corner, so this will start the border further down than the box it's a border for. I guess you could use `Rect::expanded(BorderSize)` to get the top/bottom/left/right points?

robinlinden (changes requested):
Very cool to start rendering borders!

render/render.cpp
```diff
@@ -104,6 +108,42 @@ void render_text(gfx::IPainter &painter, layout::LayoutBox const &layout, dom::T
...
+                border_size.left,
+                border_box.height - border_size.top - border_size.bottom};
+        auto border_color = style::get_property(*layout.node, "border-left-color"sv).value_or(color);
+        painter.fill_rect(border, parse_color(border_color));
+    }
+    if (border_size.right > 0) {
+        geom::Rect border{border_box.x + border_box.width - border_size.right,
```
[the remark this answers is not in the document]
mkiael: Yes, that looks better!

render/render.cpp
mkiael: Yea, doing like that makes vertical borders render correct when there is no horizontal borders.

robinlinden (approved):
Looks great!

PR #212, opened 2022-05-31
dom: Port from range-v3 to std::ranges

robinlinden:
Looks like Clang doesn't yet support libstdc++ views: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102807

libc++ ranges support is a bit wip: https://libcxx.llvm.org/Status/Ranges.html, but
`std::ranges::copy` should be supported: https://reviews.llvm.org/D122982
`std::ranges::views::{transform, filter}` should be supported: https://reviews.llvm.org/D109086 https://reviews.llvm.org/D103056
, but I guess that won't be released until Clang 15.


robinlinden:
Closing this as there isn't a Clang version with sufficient std::ranges-support yet. Will revisit after Clang 15 has been released.

PR #215, opened 2022-06-01
build: Disable rtti

robinlinden:
We're not using it yet, and since we've gotten quite far without it now
I figure we should make a conscious decision if/when to introduce it
rather than let it sneak in by accident.


mkiael (approved):
Good idea!

PR #227, opened 2022-06-04
Add new class for high level painter

mkiael:
This PR renames existing painter related classes to canvas and then adds a new painter class.
Thought canvas could be for low level stuff and the extend the painter class high level paint functions.

@robinlinden Suggestions are welcome. As always, naming is hard.


gfx/painter.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "gfx/icanvas.h"
+
+namespace gfx {
+
+class Painter {
+public:
+    Painter(ICanvas &canvas) : canvas_{canvas} {}
+
+    void draw_rect(geom::Rect const &, Color);
```
robinlinden, line 16: I called it `fill_rect` in `IPainter` as we might want a non-filling variant of it as well. :P

gfx/font.h
```diff
@@ -0,0 +1,22 @@
...
+    std::string_view font;
+};
+
+struct FontSize {
+    int px{10};
+};
+
+} // namespace gfx
+
+#endif
```
robinlinden, line 22: Missing newline at eof. I thought this was enforced by clang-format, but I guess not.. Also missing in painter.{cpp,h}

gfx/painter.h
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+#ifndef GFX_PAINTER_H_
+#define GFX_PAINTER_H_
+
+#include "gfx/icanvas.h"
+
+namespace gfx {
+
+class Painter {
```
robinlinden, line 12: Should this be header-only?

robinlinden (approved):
Looks good, with some minor cleanup-related comments. I think the naming is perfectly fine!

Would have been nice to have an example of a higher-level paint function, but that can be added in a separate PR. :P

PR #236, opened 2022-06-10
css: Add function for converting css::Rule to string

mkiael:
@robinlinden What do you think about having a widget for showing the stylesheet?


robinlinden:
It would be helpful to have a way of displaying the stylesheet(s)! :P
See: https://github.com/robinlinden/hastur/issues/34

robinlinden (approved):
Looks good!

PR #237, opened 2022-06-10
Add support for disabling tests, disable everything accessing live websites in CI

robinlinden:
@mkiael Think something like this is fine for now?

I'm hacking on a nicer abstraction that will allow testing this without a live web server to speak with, but this unblocks CI so we can continue merging things. It's been on my todo for a bit, but this makes it more important. :P


mkiael:
Looks good!

mkiael (approved):

PR #239, opened 2022-06-12
Add a IProtocolHandler abstraction

robinlinden:
This makes it possible to replace the protocol handlers in tests and add new protocols to the same API without having to keep them all in the same library.


mkiael (approved):
This looks very solid. Great work!

PR #240, opened 2022-06-13
style: Match :link and :any-link psuedo-classes

robinlinden:


style/style.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "style/style.h"
 
+#include "util/string.h"
```
mkiael, line 7: This should be in the first commit

style/BUILD
```diff
@@ -14,6 +14,7 @@ cc_library(
     deps = [
         "//css",
         "//dom",
+        "//util:string",
```
mkiael, line 17: This should be in the first commit

mkiael (changes requested):

style/style.cpp
robinlinden: Good catch! Switching the order of them was a last-minute change, so I messed up testing them. :x

PR #242, opened 2022-06-13
Stylesheet widget

### LINKED ISSUE #34: Support displaying the current style sheet(s) in the debug gui

mkiael:
Closes #34

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -4,6 +4,7 @@
 
 #include "browser/gui/app.h"
 
+#include "css/rule.h"
 #include "dom/dom.h"
 #include "gfx/opengl_canvas.h"
 #include "gfx/painter.h"
@@ -18,6 +19,7 @@
 #include <spdlog/spdlog.h>
 
 #include <functional>
+#include <sstream>
 #include <string_view>
 #include <utility>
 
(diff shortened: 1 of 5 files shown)


browser/gui/app.cpp
```diff
@@ -353,6 +366,15 @@ void App::run_dom_widget() const {
     ImGui::End();
 }
 
+void App::run_stylesheet_widget() const {
+    ImGui::SetNextWindowPos(
+            ImVec2(0, 70 * static_cast<float>(scale_) + window_.getSize().y / 2.f), ImGuiCond_FirstUseEver);
```
mkiael, line 371: Suggestions are welcome for how to handle the y-position here.

robinlinden: All of the positions and sizes here are very hacked together. :(

Maybe we can do something cleaner using the imgui tab API? Would reduce the number of windows you end up with by default too. If it works well enough, we could even switch to the tab/window docking branch of imgui if it works well enough.

robinlinden (approved):
Looks good and worked well when I tested it locally! I was missing this! :P

browser/gui/app.cpp
mkiael: Yeah, that sounds good. I'll leave this as it is for now.

PR #255, opened 2022-06-22
js: Set up a new library with a basic executable JS AST

robinlinden:
This is very incomplete, but what do you think, @mkiael?


mkiael (approved):
Exciting stuff!! Looks good!

Issue #238, opened 2022-06-11
Normalize URIs


robinlinden:
~Right now, we can't navigate to a page like `HTTP://example.com` as `//protocol` assumes a normalized scheme.~

See: https://en.wikipedia.org/wiki/URI_normalization

Zer0-One:
I think this is probably closed by #633 . The serialization process handles most of this stuff.

robinlinden:
Let's keep this open until we have tests that confirm that the normalizations that preserve semantics are implemented, and we've switched to the new url-library.

Until this is integrated with the rest of the system it's something we'll have to bear in mind when handling urls.

Zer0-One:
@robinlinden Do you think it would always be desirable to perform these normalizations automatically on serialization? Or maybe I could add a boolean parameter to `Url::serialize()` to request semantics-preserving normalization?

robinlinden:
Hmm. Maybe just adding tests to document what normalizations we/whatwg perform is enough for now, and then we can look into it again if we hit problems? If we add additional normalizations that aren't part of the whatwg url spec, they should probably be a separate thing from the spec-compliant bits.

I'm thinking that e.g. caching things internally in the browser can run into problems if `example.com/%7e/`, `example.com/%7E/`, and `example.com/~/` are interpreted as different urls, but later problem? The ones that were causing problems earlier were that there was no casing-normalization of the scheme, and that the path could be empty instead of empty being equivalent to `/` iirc.

PR #259, opened 2022-06-25
Move uri path normalization from the browser engine to the uri parser

robinlinden:
Part of #238


mkiael (approved):
Looks good! Subject line for last commit is a bit long :P

robinlinden:
> Looks good! Subject line for last commit is a bit long :P

Good catch! Didn't realize. :P

PR #261, opened 2022-06-27
Reduce the number of places where we use exceptions

robinlinden:
This is me dusting off my oldish wip `-fno-exceptions` branch. We can't quite build with it yet, but it's closer.

The main thing remaining is asio, which requires us to provide our own exception handler by defining a template function that's visible everywhere they want to throw.

### DIFF
--- a/gfx/sfml_canvas.cpp
+++ b/gfx/sfml_canvas.cpp
@@ -17,6 +17,7 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <system_error>
 
 using namespace std::literals;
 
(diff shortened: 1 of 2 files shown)


gfx/sfml_canvas.cpp
```diff
@@ -81,10 +81,13 @@ constexpr std::string_view border_fragment_shader{
         }
         )"sv};
 
-auto get_font_dir_iterator(std::filesystem::path const &path) try {
-    return std::filesystem::recursive_directory_iterator(path);
-} catch (std::filesystem::filesystem_error const &) {
-    return std::filesystem::recursive_directory_iterator();
+std::filesystem::recursive_directory_iterator get_font_dir_iterator(std::filesystem::path const &path) {
+    std::error_code errc;
```
mkiael, line 85: Should `<system_error>` be included?

mkiael (approved):

gfx/sfml_canvas.cpp
robinlinden: It would make me and include-what-you-use happier, so added it! Good catch. :P

PR #262, opened 2022-06-27
gfx: Give the shaders their own files to live in

robinlinden:
We talked about how to yoink the shaders out of the .cpp-file, and I gave one of the methods a try. :P What do you think? @mkiael 

The file endings I went with are from https://www.khronos.org/opengles/sdk/tools/Reference-Compiler/


mkiael (approved):
I think this looks great!

PR #277, opened 2022-08-02
net: Simplify initialization and read_bytes

robinlinden:


mkiael (approved):
That looks much better!

PR #283, opened 2022-08-09
os: Allow overriding the default scale

robinlinden:
This patch makes it possible to set the default scale by setting the
environment variable `HST_SCALE`.

### DIFF
--- a/os/linux.cpp
+++ b/os/linux.cpp
@@ -32,9 +32,9 @@ std::vector<std::string> font_paths() {
 }
 
 unsigned active_window_scale_factor() {
-    // Qt, Gnome, and Elementary in that order.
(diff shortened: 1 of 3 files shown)


os/linux.cpp
```diff
@@ -8,6 +8,8 @@
 #include <charconv>
 #include <cstdlib>
 #include <cstring>
+#include <optional>
+#include <string_view>
```
mkiael, line 12: Are these includes needed?

mkiael (changes requested):

os/linux.cpp
robinlinden: Nope! Removed them! This was originally a lot more complicated than it is now, and I just missed nuking the bonus includes. :P Good catch!

mkiael (approved):

PR #292, opened 2022-08-19
uri: Add normalization of casing of the scheme and host

robinlinden:
Part of #238

### DIFF
--- a/uri/uri.cpp
+++ b/uri/uri.cpp
@@ -5,6 +5,8 @@
 
 #include "uri/uri.h"
 
+#include "util/string.h"
+
 #include <regex>
 #include <utility>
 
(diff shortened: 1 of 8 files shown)


uri/uri.cpp
```diff
@@ -13,6 +15,16 @@ namespace {
 
 // https://en.wikipedia.org/wiki/URI_normalization#Normalization_process
 void normalize(Uri &uri) {
+    // The scheme and host components of the URI are case-insensitive and
+    // therefore should be normalized to lowercase.
+    std::transform(begin(uri.scheme), end(uri.scheme), begin(uri.scheme), [](unsigned char c) {
+        return static_cast<char>(std::tolower(c));
+    });
```
mkiael, line 22: Maybe we could have a `std::string to_lower(std::string)` in `util/string`?

robinlinden: Good idea! Will add. :P

robinlinden: I added it, but I'm not sure if I prefer `string lowercased(string)` more than `string to_lower(string)`. It follows the naming convention we use in `//geom` for the methods that return a new instance, and I feel like the name is less ambiguous. That said, `to_lower` is symmetrical with what we're doing with `char` and what's in the stdlib, and we prevent accidental misuse with `[[nodiscard]]`. :P

mkiael (approved):

PR #295, opened 2022-08-21
Add locale-independent character helpers to `//util:string`

robinlinden:
We had a bunch of issues if the locale wasn't the default `C` one, so I replaced the sort of error-prone `<cctype>` functions with the constexpr versions of them we were already using in `html2`.


mkiael (approved):
Looks great!

PR #296, opened 2022-08-22
dom: Reformat tests in a way that clang-format tolerates

robinlinden:


robinlinden:
1.5 or so years later, but this PR is ready now that MSVC no longer crashes when trying to compile things like `std::variant` or `{.designated_initializers}` quite as often.

PR #302, opened 2022-08-27
Update README with compiler versions

mkiael:


mkiael:
Should I prefix the commit message with something?

robinlinden:
We still test against gcc-10 in the coverage build, so that's why it's still in there. :P I've been prefixing documentation changes with `doc:`.

We could drop gcc 10, but I haven't bothered since anything I've been missing in gcc-10 is also missing in clang-14, so dropping support wouldn't give us anything aside from removing the gcc-10-specific hack in `util/overloaded.h`.

PR #303, opened 2022-08-27
build: Disable warnings from dependencies

mkiael:
I had to disable these warnings when building with gcc 12.0.1. A bit strange since this is the same version that we use in our ci build?


mkiael:
Never mind. I forgot to update my `.bazelrc.local` file.

robinlinden:
Yeah, it's a bit unfortunate. :( Those warnings aren't supported in Clang yet, so GCC requires `--config gcc12` right now. Hopefully we can clean this up in the future.

PR #304, opened 2022-08-27
browser/gui: Add support for simple navigation history

robinlinden:
Currently set to `alt+left` and `alt+right`.


mkiael (approved):
Very nice!

PR #305, opened 2022-08-28
Add support for rendering borders with radius

mkiael:
This is a first step in adding support for rendering borders with radius.

We are still missing support for parsing the border-radius property in our CSS parser, so we will not be able to see this in action yet.

Part #99


mkiael:
I should probably add some tests since we got that setup for `render.cpp` now.

gfx/rect_shader.frag
```diff
@@ -0,0 +1,200 @@
...
+        // Otherwise, set transparent color
+        } else {
+            gl_FragColor = no_color;
+        }
+    }
+    // Otherwise, set transparent color
+    else {
+        gl_FragColor = no_color;
+    }
+}
```
robinlinden, line 200: Missing newline at EOF.

robinlinden (approved):
Looks great! Really good job. :)
Feel free to skip adding tests for render unless you want to. You didn't touch it much, and the only test for it is my example. :P

robinlinden (approved):
Nice tests too!

PR #308, opened 2022-08-29
Target C++23

robinlinden:
Now that we have a few C++23-features in all major compilers, I figure it's time we start playing around with them. :P
I'm working on some style property features that make use of [Monadic operations for `std::optional`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0798r8.html), and switching to C++23 feels like something major and separate from upcoming style property PR.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,7 +20,7 @@ build --incompatible_enable_cc_toolchain_resolution
 # not relying on transitive dependencies, so let's enable that!
 build:linux --features=layering_check
 
-build:linux --cxxopt='-std=c++20'
+build:linux --cxxopt='-std=c++2b'
 build:linux --cxxopt='-fno-rtti'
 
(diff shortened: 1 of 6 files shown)


robinlinden:
This is on hold while we look into code coverage with gcc 12.

Zer0-One:
```
INFO: Elapsed time: 49.239s, Critical Path: 10.75s
INFO: 494 processes: 22 internal, 472 linux-sandbox.
INFO: Build completed successfully, 494 total actions
```
```
zero-one@proton ~/repos/hastur $ clang --version
clang version 15.0.7
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm/15/bin
Configuration file: /etc/clang/clang.cfg
```

LGTM

mkiael (approved):

PR #310, opened 2022-08-30
render: Fix fully transparent boxes being rendered

robinlinden:
Without this patch, we only ignore fully transparent boxes if they also
match the render-internal magical transparent color constant.


mkiael (approved):
Great!

PR #316, opened 2022-09-04
css2: Handle comments during tokenization

mkiael:
Trying out a similar tokenizer for css2 as for html2. Some code duplication here now, but I guess we could make a base class for the tokenizer to reduce it.

Part of #31


robinlinden (approved):
Looks good! Maybe add some links to the spec though? :P Also, when bringing code over from other files, try to bring the copyright with it.

Issue #19, opened 2021-09-18
Expand CSS shorthand properties


robinlinden:
See: https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties

Expanding these while parsing/adding them to the style sheets will make things like
```
padding: 10px;
padding-left: 20px;
```
work correctly and make things in general a lot more convenient.

## Progress
- [ ] Background properties
- [x] Font properties
- [x] Border properties
- [x] Margin properties
- [x] Padding properties

robinlinden:
Let's close this and create more specific issues if we need and/or feel like it. We've got most of it done in most cases.

PR #324, opened 2022-09-08
Add very partial support for the background shorthand property

robinlinden:
When working on my other css property work I noticed a website I expected to have a gray background was white even though we support the background-color property. Turns out there's a background shorthand property that a lot of people use.


mkiael (approved):

robinlinden:
Part of #19

PR #329, opened 2022-09-12
Switch to std::ranges algorithm variants where possible

robinlinden:


robinlinden:
Apparently "complete" doesn't mean complete-complete: https://releases.llvm.org/15.0.0/projects/libcxx/docs/Status/Cxx20.html#:~:text=P0896R4
libc++ trunk appears to have `std::ranges::max_element`, so it'll be fixed in the next release, I guess.

15: https://godbolt.org/z/of89o5Ybb
trunk: https://godbolt.org/z/zorEcjsEa

mkiael (approved):

PR #338, opened 2022-09-19
build: Drop untested gcc11 config

robinlinden:
Ubuntu 20.04 only has up to gcc 10, and Ubuntu 22.04 skips past gcc 11 to also support gcc 12, so supporting gcc 11 doesn't really get us anything.


robinlinden:
We could add CI for this if we want to keep it, but that seems like more effort than it's worth given that we have access to newer compilers on the Ubuntu 22.04 LTS.

mkiael (approved):

PR #349, opened 2022-09-29
Nuke range-v3 dependency

robinlinden:
In a way this is a bit sad since std::ranges is a lot nicer than what we do right now, but it's also a dependency, and hopefully Clang 16 will finish their range implementation. Even updating to Clang 15 would allow us to use std::ranges for almost all <algorithm> things we do right now.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -118,13 +118,6 @@ http_archive(
 )
 
 http_archive(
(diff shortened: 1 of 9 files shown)


mkiael (approved):
Good move. Our usage did not really justify pulling in that dependency.

PR #355, opened 2022-10-06
html2: Fix the replacement character not being emitted in UTF-8

### LINKED ISSUE #135: The replacement character should be emitted in utf8, not utf16
Tests need to be added and this needs to be changed to be emitted as `0xEF 0xBF 0xBD` instead of `0xFF 0xFD`.

See: https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character

robinlinden:
Resolves #135


mkiael (approved):
Nice increase of code coverage!

PR #357, opened 2022-10-11
css2: Add partial support for tokenizing strings

mkiael:
Implementation is based on this chapter in the specification: https://www.w3.org/TR/css-syntax-3/#consume-string-token

### DIFF
--- a/css2/tokenizer_test.cpp
+++ b/css2/tokenizer_test.cpp
@@ -135,5 +135,33 @@ int main() {
         expect_token(output, DelimToken{'a'});
         expect_token(output, WhitespaceToken{});
     });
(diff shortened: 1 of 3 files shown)


robinlinden (approved):
Neat!

css2/tokenizer_test.cpp
```diff
@@ -135,5 +135,33 @@ int main() {
         expect_token(output, DelimToken{'a'});
         expect_token(output, WhitespaceToken{});
     });
+
+    etest::test("single qouted string", [] {
```
robinlinden, line 139: qouted -> quoted

PR #361, opened 2022-10-14
etest: Support custom assertion failure messages

### LINKED ISSUE #268: Allow specifying failure messages for expect/require-calls
Right now, every expect/require failure use the same error message: `expectation failure at gui/window_test.cpp(50:15)`
```
expect(window.children.empty(), std::format("Unexpected widgets {}", window.children));
```
should probably print something like `gui/window_test.cpp(50:15): Unexpected widgets {"url_bar"}`
instead of `expectation failure at gui/window_test.cpp(50:15)`

Not 100% sure about what kind of API I like for this.
```
expect(window.children.empty(), std::format("Unexpected widgets {}", window.children));
expect(window.children.empty()).on_fail(std::format("Unexpected widgets {}", window.children));
expect(window.children.empty()) << std::format("Unexpected widgets {}", window.children);
expect(window.children.empty(), {.on_fail = std::format("Unexpected widgets {}", window.children)});
```

robinlinden:
Resolves #268

### DIFF
--- a/css/parser_test.cpp
+++ b/css/parser_test.cpp
@@ -73,7 +73,7 @@ template<class KeyT, class ValueT>
 ValueT get_and_erase(std::map<KeyT, ValueT, std::less<>> &map,
         KeyT key,
         etest::source_location const &loc = etest::source_location::current()) {
-    require(map.contains(key), loc);
+    require(map.contains(key), {}, loc);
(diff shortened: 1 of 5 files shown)


mkiael (approved):
Nice!

PR #363, opened 2022-10-15
css2: Add initial support for ident token

mkiael:
Support for escaped characters are still missing.


robinlinden (approved):
Looks good! Great job!

PR #369, opened 2022-10-18
Move handling of origin/scheme-relative uris into `//uri`

robinlinden:
Previously this was sort of handled in 2 different places in `//engine`, and would have had to be added to yet more places. This also gives us one place to add support for e.g. relative uris which was what I was originally planning on doing.

@mkiael Not sure what to call this member function. `Uri Uri::complete_uri(std::string) const` doesn't feel super intuitive. Any ideas? One thing I considered was adding a `std::optional<Uri> const &base` argument to `Uri::parse`.

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -294,7 +294,7 @@ int App::run() {
 
 void App::navigate() {
     page_loaded_ = false;
-    auto uri = uri::Uri::parse(url_buf_);
+    auto uri = uri::Uri::parse(url_buf_, engine_.uri());
     browse_history_.push(uri);
     engine_.navigate(std::move(uri));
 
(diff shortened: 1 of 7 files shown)


mkiael:
Nice cleanup! Hard to come up with a good name. Maybe `join` or `join_uri`?

mkiael (approved):

robinlinden:
> Nice cleanup! Hard to come up with a good name. Maybe `join` or `join_uri`?

Not sure? Part of it is that sometimes the base uri doesn't affect the new uri at all.. What do you think of just adding an optional base_uri argument to `Uri::parse`? I read through part of the whatwg url spec, and it sounds like that's what they do there: https://url.spec.whatwg.org/#concept-url-parser

mkiael:
Ah, ok. Then I think we should try that!

robinlinden:
@mkiael Please take another look. Updated with passing the optional base_uri to Uri::parse now. :P

mkiael (approved):
Looks great!

PR #371, opened 2022-10-20
layout: Support fractional sizes in pixel calculation

robinlinden:
The main things being that now we support fractional sizes, and warn on things we don't handle.

### DIFF
--- a/layout/layout.cpp
+++ b/layout/layout.cpp
@@ -87,6 +87,11 @@ std::map<std::string_view, float> const kFontSizeAbsoluteSizeKeywords{
 // * Not all measurements have to be in pixels.
 // * %, rem
 int to_px(std::string_view property, int const font_size) {
+    // Special case for 0 since it won't ever have a unit that needs to be handled.
+    if (property == "0") {
+        return 0;
+    }
+
     int res{};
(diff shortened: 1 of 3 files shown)


layout/layout.cpp
```diff
@@ -99,10 +99,19 @@ int to_px(std::string_view property, int const font_size) {
         return res;
     }
 
-    if (property.ends_with("em")) {
+    auto const parsed_length = std::distance(property.data(), parse_result.ptr);
+    auto const suffix = property.substr(parsed_length);
```
mkiael, line 103: I think they call this `unit` in the CSS spec.

mkiael (approved):

layout/layout.cpp
robinlinden: Good catch! Will update the name to match the spec. Also need to add a workaround for gcc 10 not supporting from_chars with floats. :(

robinlinden: Right, renamed it and added a very not spec-compliant version of from_chars for gcc 10. Can you please have a look at the last commit again?

robinlinden: Also added a few tests to make sure we run a few more of the failure paths in the layout system.

mkiael: Looks good!

PR #378, opened 2022-10-24
css2: Simplify identification of ident sequence

mkiael:
This commits adds a function for identifying if a sequnce of character starts an ident sequence. Support for handling escaped characters is still missing.


robinlinden (approved):
Nice cleanup!

Some small typos in the commit message:
`...commits adds a function for identifying if a sequnce of character...` -> 
`...commit adds a function for identifying if a sequence of characters...`

PR #386, opened 2022-11-08
css: Set up a more strongly-typed and less error-prone CSS property ID abstraction

robinlinden:


robinlinden:
@mkiael The big and annoying migration commit should mostly be a find-and-replace from the strings to the enum version.

If you want, I can sink some time into splitting it into 1 commit per layer w/ a call to `property_id_from_string` as a translation layer. The migration from my side was mostly `sed`ding a whole bunch and it ended up being more convenient to `sed` it all at once. :P

css/property_id_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "css/property_id.h"
+
+#include "etest/etest.h"
+
+#include <fmt/format.h>
+
+#include <string_view>
+#include <utility>
```
mkiael, line 12: Is this include needed?

mkiael (approved):
This is great! Good job!

css/property_id_test.cpp
robinlinden: I'm sure it was at one point, but it's gone now. :P Good catch.

PR #388, opened 2022-11-09
js: Add support for native functions

robinlinden:
This enables us to expose browser functionality and start writing a JS platform library in a very simple way.

E.g. a very simplified version of [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console/log):
```
std::ostream &log_handle = browser.gui_log_handle();

AstExecutor e;
e.variables["console.log"] = NativeFunction{[&](auto const &args) {
  for (auto const &arg : args) { log_handle << arg << '\n'; }
  return Value::undefined();
}};
```

### DIFF
--- a/js/ast.h
+++ b/js/ast.h
@@ -5,6 +5,7 @@
 #ifndef JS_AST_H_
 #define JS_AST_H_
 
+#include <functional>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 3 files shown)


mkiael (approved):
Good idea! Looks great.

PR #390, opened 2022-11-11
layout: Get rid of duplicate font size constant

robinlinden:


robinlinden:
The commit should obviously say font size and not pixel size. Will fix. :P Guess that's what I get for working late.

mkiael (approved):

PR #391, opened 2022-11-12
Add more utility functions to string.h

Zer0-One:
This is preparation for a new URL parser.

### DIFF
--- a/util/base_parser.h
+++ b/util/base_parser.h
@@ -1,12 +1,13 @@
 // SPDX-FileCopyrightText: 2021 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2022 David Zero <zero-one@zer0-one.net>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
 #ifndef UTIL_BASE_PARSER_H_
 #define UTIL_BASE_PARSER_H_
 
-#include <algorithm>
-#include <array>
+#include "util/string.h"
+
 #include <concepts>
 #include <cstddef>
 #include <string_view>
@@ -32,6 +33,15 @@ public:
 
     constexpr void advance(std::size_t n) { pos_ += n; }
 
+    constexpr void back(std::size_t n) { pos_ -= n; }
+
+    constexpr void reset() { pos_ = 0; }
+
+    constexpr void reset(std::string_view input) {
+        input_ = input;
+        pos_ = 0;
+    }
+
     template<Predicate T>
     constexpr std::string_view consume_while(T const &pred) {
         std::size_t start = pos_;
@@ -42,15 +52,12 @@ public:
     }
 
     constexpr void skip_whitespace() {
-        while (!is_eof() && is_space(peek())) {
+        while (!is_eof() && util::is_whitespace(peek())) {
             advance(1);
         }
     }
 
 private:
-    static constexpr auto space_chars = std::array{' ', '\f', '\n', '\r', '\t', '\v'};
-    constexpr static bool is_space(char c) { return std::ranges::find(space_chars, c) != end(space_chars); }
-
     std::string_view input_;
     std::size_t pos_{0};
 };
--- a/util/BUILD
+++ b/util/BUILD
@@ -1,9 +1,17 @@
 load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
 
+dependencies = {
+    "base_parser": [":string"],
+}
+
 [cc_library(
     name = hdr[:-2],
     hdrs = [hdr],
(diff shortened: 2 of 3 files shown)


util/base_parser.h
```diff
@@ -32,6 +32,15 @@ class BaseParser {
 
     constexpr void advance(std::size_t n) { pos_ += n; }
 
+    constexpr void back(std::size_t n) { pos_ -= n; }
+
+    constexpr void reset() { pos_ = 0; }
+
+    constexpr void reset(std::string_view input) {
```
robinlinden, line 39: What's the use-case for these reset and back member functions? So far we've only used this for throwaway parsers, e.g.
`CssStyleSheet css = CssParser::parse("html { font-size: 1em; }");` where `parse(char const *)` internally creates a parser, creates a stylesheet, and returns it.

util/base_parser.h
```diff
@@ -42,15 +51,32 @@ class BaseParser {
...
+        while (!is_eof() && is_whitespace(peek())) {
             advance(1);
         }
     }
 
-private:
-    static constexpr auto space_chars = std::array{' ', '\f', '\n', '\r', '\t', '\v'};
-    constexpr static bool is_space(char c) { return std::ranges::find(space_chars, c) != end(space_chars); }
+    static constexpr bool is_whitespace(char c) {
+        return c == 0x20 || c == 0x0c || c == 0x0a || c == 0x0d || c == 0x09 || c == 0x0b;
```
robinlinden, line 60: Why change these from `'\t'` and friends to ascii hex? imo it's more readable as `'\n'` when we're talking about printable characters anyway.

Especially for things like `is_upper_alpha`. Why not `return c >= 'A' && c <= 'Z';`?

robinlinden (changes requested):
Can you please update the commit message to follow convention (`target: change`)? E.g. `util/base_parser: Add more utility functions`

Also feel free to add yourself to the copyright header in `base_parser.h`.

util/base_parser.h
Zer0-One: There's plenty of places in the WHATWG URL standard where they describe the parser going backward. Ctrl+F "decrease pointer". Same with starting over from the beginning of the parser input.

util/base_parser.h
Zer0-One: I was trying to keep it consistent. I can use escape characters instead for everything.

util/base_parser.h
robinlinden: Starting over from the beginning is sad, but fair enough, that's how the spec goes. :P

I'm mostly wondering about `reset(std::string_view)`, but I guess maybe you create a `std::string` for the `UrlParser` to keep alive and pass that to reset?

util/base_parser.h
robinlinden: Yes, please. Printable characters being readable without an ascii table would be nicer imo.

util/base_parser.h
Zer0-One: Yup, exactly.

util/base_parser.h
```diff
@@ -42,15 +52,32 @@ class BaseParser {
...
+
+    static constexpr bool is_upper_alpha(char c) { return c >= 'A' && c <= 'Z'; }
+
+    static constexpr bool is_lower_alpha(char c) { return c >= 'a' && c <= 'z'; }
+
+    static constexpr bool is_digit(char c) { return c >= '0' && c <= '9'; }
+
+    static constexpr bool is_alpha(char c) { return is_upper_alpha(c) || is_lower_alpha(c); }
+
+    static constexpr bool is_alphanumeric(char c) { return is_alpha(c) || is_digit(c); }
```
mkiael, line 78: Some of these methods are already available in `util/string.h`.

Zer0-One: I moved them to string.h, and removed the duplicate definition of is_whitespace() from BaseParser.

util/base_parser.h
```diff
@@ -1,10 +1,13 @@
...
+// SPDX-FileCopyrightText: 2022 David Zero <zero-one@zer0-one.net>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
 #ifndef UTIL_BASE_PARSER_H_
 #define UTIL_BASE_PARSER_H_
 
+#include "util/string.h"
+
 #include <algorithm>
```
robinlinden, line 11: This doesn't need `<algorithm>` or `<array>` now that you've removed `is_space`.

robinlinden (changes requested):

robinlinden (approved):

mkiael (approved):

PR #397, opened 2022-11-19
build: Rewamp how we handle warning flags

robinlinden:
This should make updating 3rd-party deps a bit more convenient at the expense of knowing less about the quality of the libraries we pull in.


robinlinden:
@mkiael What do you think about managing our warnings like this instead?

mkiael:
I think it looks good. Nice work!

mkiael (approved):

PR #403, opened 2022-12-01
html2: Support tokenizing doctype system identifiers

### LINKED ISSUE #207: Support tokenizing public and system doctype identifiers

robinlinden:
With this change, we can now load msn.com without crashing.

Resolves #207


mkiael (approved):
Nice work!

PR #412, opened 2022-12-07
Reverse the font search path to look in system first.

GrayHatter:
do you really want this? ¯\_(ツ)_/¯ I didn't give it that much thought, but it make text render for me :)


robinlinden:
@GrayHatter This isn't needed now that we're a bit better with the font-family handling, right? User-specified paths getting priority over the system paths seems like a reasonable way to go about things. :P

PR #424, opened 2022-12-20
deps: Simplify depending on GitHub archives

robinlinden:
@mkiael What do you think about something like this? I wanted to reduce duplication, but I'm not sure if this helps enough to make it worth it.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -11,9 +11,9 @@ http_archive(
 
 http_archive(
     name = "rules_fuzzing",
-    sha256 = "d9002dd3cd6437017f08593124fdd1b13b3473c7b929ceb0e60d317cb9346118",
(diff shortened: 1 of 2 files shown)


mkiael (approved):
Nice. I think this makes it cleaner.

Zer0-One:
Personally I don't think it makes enough of a difference to bother adding. Just more complexity.

robinlinden:
I liked the idea of being able to replace
```
    strip_prefix = "rules_fuzzing-0.3.2",
    url = "https://github.com/bazelbuild/rules_fuzzing/archive/v0.3.2.zip",
```
w/
```
    repo = "bazelbuild/rules_fuzzing",
    version = "v0.3.2",
```
and dropping the build-file arg and just use ours if we had one, but turns out switching on if a file exists in Starlark was non-trivial and projects on GitHub don't always follow reasonable conventions. Closing this for now, but might revisit if there's a nice way of reducing the duplication when declaring dependencies.

PR #436, opened 2023-01-06
gfx+render: Use the first available font family instead of only trying the first one

robinlinden:


GrayHatter:
can confirm, is fixed

PR #443, opened 2023-01-12
all: Restrict the fuzz tests to only run when using clang

robinlinden:
No idea why it was set up this way when no other fuzz test was. Past me didn't document it either. :(

### DIFF
--- a/bzl/BUILD
+++ b/bzl/BUILD
@@ -5,6 +5,11 @@ filegroup(
 )
 
 config_setting(
+    name = "is_clang",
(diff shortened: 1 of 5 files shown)


robinlinden:
Right, I guess there was a reason past me didn't run it for gcc. Unable to reproduce the failure locally, and it seems to pass with gcc 10 though, so will investigate further.

robinlinden:
So after reading the rules_fuzzing docs, this PR took a turn. Now we always block running fuzz tests using gcc as that's not supported at all by rules_fuzzing.

mkiael (approved):

PR #444, opened 2023-01-15
net/ip: Add IPv4 and IPv6 address serialization

Zer0-One:
This is more prep for generating blob URLs.


net/ip.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef NET_IP_H_
+#define NET_IP_H_
+
+#include <cstdint>
+#include <string>
+#include <vector>
+
+namespace net {
+
+struct IP4Address {
```
robinlinden, line 14: Do these need to be classes? As they are right now, they feel like a roundabout way of doing something like `std::string as_ipv4_string(std::uint32_t);`?

net/ip.cpp
```diff
@@ -0,0 +1,85 @@
+// SPDX-FileCopyrightText: 2023 David Zero <zero-one@zer0-one.net>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <cmath>
+#include <cstdint>
+#include <string.h>
```
robinlinden, line 7: Prefer the C++ header variants. `<cstring>` instead of `<string.h>`. Also make sure you prefix things from the stdlib with `std::` as there's no guarantee that the implementation dumps things into the global namespace when using the C++ headers.

net/ip.h
```diff
@@ -0,0 +1,38 @@
...
+    IP4Address(uint32_t address) { addr = address; }
+
+    uint32_t addr;
+
+    std::string serialize(void);
+};
+
+struct IP6Address {
+    IP6Address(std::vector<uint16_t> address) {
```
robinlinden, line 23: How about something like `std::span<std::uint16_t, 8>` or `std::array<std::uint16_t, 8>` instead? Neither would require heap allocations the way a std::vector does and it would also enforce from the call-site that you pass something with exactly 8 elements, so you don't have the silently doing the wrong thing and reading from uninitialized memory the current implementation has.

robinlinden (changes requested):

net/ip.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::uint32_t n = addr;
+
+    for (std::size_t i = 1; i < 5; i++) {
+        out.insert(0, std::to_string(n % 256));
+
+        if (i != 4) {
+            out.insert(0, ".");
+        }
+
+        n = (std::uint32_t)floor(n / 256.0);
```
robinlinden, line 27: Is the u32 -> double -> floor -> u32 dance necessary? Why not just something like `n /= 256`?

net/ip.cpp
```diff
@@ -0,0 +1,86 @@
...
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <span>
+
+#include <fmt/core.h>
+
+#include "ip.h"
```
robinlinden, line 12: `net/ip.h`. We use the full path everywhere else, and the header order (that will be enforced by clang-tidy soon(tm)) is
1. main header (the header being tested, or the header the .cpp is for)
2. our other headers
3. 3rd-party
4. stdlib

robinlinden (changes requested):
Last nitpicky comments, I promise. Looks good!

net/ip.cpp
GrayHatter: n >> 8

robinlinden (approved):

PR #447, opened 2023-01-17
Fix text not being correctly handled when scale != 1

robinlinden:


geom/geom.h
```diff
@@ -12,6 +12,15 @@ namespace geom {
 struct Position {
     int x{}, y{};
     [[nodiscard]] bool operator==(Position const &) const = default;
+
+    [[nodiscard]] constexpr Position scaled(unsigned scale, Position origin = {0, 0}) const {
+        return Position{
```
GrayHatter, line 17: `return Position{`
vs 
`return Position {`?

is this correct, or code format enforcement when?

GrayHatter: nevermind, I didn't realize you were writing logic in geom.h because geom.c doesn't exist :<

robinlinden: `Position{` is correct and how clang-format formats the code right now. :P `constexpr` requires the code to be visible at the call site, so simple things go in the header and are marked constexpr.

GrayHatter (approved):

mkiael (approved):

PR #448, opened 2023-01-18
uri/uri: Added blob URL generation function

Zer0-One:
Also adds a skeleton for a new URL class which will do proper parsing of URLs.

### DIFF
--- a/net/ip.cpp
+++ b/net/ip.cpp
@@ -13,6 +13,7 @@
 
 namespace net {
 
+// https://url.spec.whatwg.org/#concept-ipv4-serializer
 std::string ipv4_serialize(std::uint32_t addr) {
     std::string out = "";
     std::uint32_t n = addr;
@@ -30,6 +31,7 @@ std::string ipv4_serialize(std::uint32_t addr) {
     return out;
 }
 
+// https://url.spec.whatwg.org/#concept-ipv6-serializer
 std::string ipv6_serialize(std::span<std::uint16_t, 8> addr) {
     std::string out = "";
 
--- a/uri/uri.cpp
+++ b/uri/uri.cpp
@@ -5,12 +5,16 @@
 
 #include "uri/uri.h"
 
+#include "net/ip.h"
 #include "util/string.h"
+#include "util/uuid.h"
 
 #include <ctre.hpp>
 #include <fmt/format.h>
 
 #include <exception>
+#include <optional>
+#include <string>
 #include <utility>
 
 namespace uri {
@@ -95,6 +99,65 @@ Uri Uri::parse(std::string uristr, std::optional<std::reference_wrapper<Uri cons
     }
 
     return uri;
+}
+
+// https://w3c.github.io/FileAPI/#unicodeBlobURL
+std::string blob_url_create(Origin *origin) {
+    if (origin == NULL) {
+        return "";
+    }
+
+    std::string result = "blob:";
+    std::string serialized = "";
+
+    // https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin
+    if (origin->opaque) {
+        serialized = "null";
+    } else {
+        serialized = origin->scheme + "://";
+
+        switch (origin->host->type) {
+            case DNS_DOMAIN:
+            case OPAQUE:
+            case EMPTY:
+                serialized += origin->host->domain;
+                break;
+            case IP4_ADDR:
+                serialized += net::ipv4_serialize(origin->host->ip4_addr);
+                break;
+            case IP6_ADDR:
+                serialized += "[" + net::ipv6_serialize(origin->host->ip6_addr) + "]";
+        }
+
+        if (origin->port.has_value()) {
+            serialized += ":" + std::to_string(origin->port.value());
+        }
+    }
+
+    result += serialized;
+    result += "/";
+    result += util::new_uuid();
+
+    return result;
+}
+
+bool URL::parse(std::optional<URL> base) {
+    bool ret = parse_basic(base, std::nullopt, std::nullopt);
+
+    if (ret) {
+        return true;
+    }
+
+    return true;
+}
+
+ParserState URL::parse_basic(
+        std::optional<URL> base, std::optional<URL> url, std::optional<ParserState> state_override) {
+    if (base || url || state_override) {
+        return FAILURE;
+    }
+
+    return FAILURE;
 }
 
 } // namespace uri
--- a/uri/uri.h
+++ b/uri/uri.h
@@ -6,8 +6,12 @@
 #ifndef URI_URI_H_
 #define URI_URI_H_
 
+#include "util/base_parser.h"
+
 #include <functional>
+#include <map>
 #include <optional>
+#include <span>
 #include <string>
 
 namespace uri {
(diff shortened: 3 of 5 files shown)


robinlinden (changes requested):
Can you add some comments linking to the relevant spec sections implemented? See e.g. html2/tree_constructor.cpp for an example where we do this to make reviewing and maintaining the more spec-compliant implementations easier.

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
...
+
+inline const std::map<std::string, std::string> special_schemes = {
+        {"ftp", "21"}, {"file", ""}, {"http", "80"}, {"https", "443"}, {"ws", "80"}, {"wss", "443"}};
+
+/**
+ * Generates a new Blob URL for the given origin
+ */
+std::string blob_url_create(Origin *origin);
+
+class URL final : util::BaseParser {
```
robinlinden, line 98: `Url` instead of `URL` to follow naming conventions.

I think we should keep the Url and the UrlParser as separate classes. Having a Url be a Parser feels like very weird design?

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
...
+    std::string scheme;
+    Host *host;
+    std::optional<uint16_t> port;
+    std::optional<std::string> domain;
+    // Need this placeholder until I figure out what "opaqueness" means for an origin in this context
+    bool opaque;
+};
+
+enum ParserState {
+    SCHEME_START,
```
robinlinden, line 65: `SchemeStart, Scheme, NoScheme,` etc to follow naming conventions.

uri/uri.cpp
```diff
@@ -97,4 +101,63 @@ Uri Uri::parse(std::string uristr, std::optional<std::reference_wrapper<Uri cons
     return uri;
 }
 
+// https://w3c.github.io/FileAPI/#unicodeBlobURL
+std::string blob_url_create(Origin *origin) {
```
robinlinden, line 105: Why a mutable pointer? You don't change it, and the spec makes it sound mandatory, so `Origin const &`.

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
...
+    // union {
+    std::string domain;
+    std::uint32_t ip4_addr;
+    std::uint16_t ip6_addr[8];
+    //};
+};
+
+struct Origin {
+    std::string scheme;
+    Host *host;
```
robinlinden, line 57: Why does this have to be a pointer?

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
     [[nodiscard]] bool operator==(Uri const &) const = default;
 };
 
+enum HostType { DNS_DOMAIN, IP4_ADDR, IP6_ADDR, OPAQUE, EMPTY };
+
+struct Host {
+    HostType type;
+
+    // union {
```
robinlinden, line 48: ?

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
...
+    PATH_START,
+    PATH,
+    OPAQUE_PATH,
+    QUERY,
+    FRAGMENT,
+    FAILURE,
+    TERMINATE
+};
+
+inline const std::map<std::string, std::string> special_schemes = {
```
robinlinden, line 90: It feels like the URL class and friends isn't really ready to be merged? Lots of random work-in-progress bits, but no real implementation at all and no tests. It's not really possible to give any form of reasonable review when not even a basic vertical slice or at least a spec link showing that the data structure is reasonable is there.

uri/uri_test.cpp
```diff
@@ -178,5 +182,34 @@ int main() {
...
+    etest::test("blob URL generation", [] {
+        std::string REGEX_UUID = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}";
+
+        uri::Host h = {uri::DNS_DOMAIN, "example.com"};
+        uri::Origin o = {"https", &h, (std::uint16_t)8080, std::nullopt, false};
+
+        std::string blob = uri::blob_url_create(&o);
+        std::cout << "Generated Blob URL: " << blob << std::endl;
+
+        etest::require(std::regex_match(blob, std::regex("blob:https://example.com:8080/" + REGEX_UUID)));
```
robinlinden, line 194: In general, use `expect` when you don't require the test to bail on this line if this assertion fails.

uri/uri.h
```diff
@@ -36,6 +40,87 @@ struct Uri {
...
+    std::string domain;
+    std::uint32_t ip4_addr;
+    std::uint16_t ip6_addr[8];
+    //};
+};
+
+struct Origin {
+    std::string scheme;
+    Host *host;
+    std::optional<uint16_t> port;
```
robinlinden, line 58: Missing std:: namespace.

uri/uri_test.cpp
```diff
@@ -178,5 +182,34 @@ int main() {
         expect_eq(completed, uri::Uri::parse("hax://example.com/test/hello"));
     });
 
+    etest::test("blob URL generation", [] {
+        std::string REGEX_UUID = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}";
+
+        uri::Host h = {uri::DNS_DOMAIN, "example.com"};
+        uri::Origin o = {"https", &h, (std::uint16_t)8080, std::nullopt, false};
```
robinlinden, line 189: Do you really need to cast 8080 to u16 here? If so, use either a static_cast or `std::uint16_t{8080}`.

robinlinden (changes requested):
Let's build all the new `Url` stuff in in a new top-level `url` folder instead of messing with the `Uri` bits that should get nuked once the new stuff is in place.

The blob generation and the parser should be separate commits as well.

uri/uri.cpp
```diff
@@ -97,4 +101,63 @@ Uri Uri::parse(std::string uristr, std::optional<std::reference_wrapper<Uri cons
     return uri;
 }
 
+// https://w3c.github.io/FileAPI/#unicodeBlobURL
```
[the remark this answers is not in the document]
Zer0-One: The URL spec is coupled to the File API. The spec says that a URL has an associated blob URL entry. I saw that, and implemented blob URL generation as a separate thing, because two the concepts are related, and because I figure generating URLs should be a feature of a URL library. It's not part of URL parsing. I can get rid of it if we don't want it.

uri/uri.h
Zer0-One: Those should be a union, but I don't know enough C++ to know how to do that idiomatically. Unions only work the way i expect with trivial types.

uri/uri_test.cpp
Zer0-One: > Do you really need to cast 8080 to u16 here?

IIRC, yes, otherwise msvc complains.

uri/uri.h
Zer0-One: idk, it seems ok to me, but I guess it would be fine to split them out.

net/ip.cpp
```diff
@@ -13,6 +13,7 @@
 
 namespace net {
 
+// https://url.spec.whatwg.org/#concept-ipv4-serializer
```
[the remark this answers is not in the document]
Zer0-One: It's 2 comment lines. Does it really deserve its own commit?

uri/uri.h
robinlinden: Ah, right. Yeah, unions has a bunch of requirements regarding construction/destruction and special members. Usually `std::variant<Foo, Bar, Baz>` is used instead.

uri/uri.h
robinlinden: We can also discuss design on Tox or IRC if you want. I can't tell you if any of what's in here is reasonable just looking at what we have right now. The API you call to start parsing a url looks like what I started with before you wanted to take over. Very spec-compliant. :P

PR #449, opened 2023-01-20
 js: Add support for while statements 

robinlinden:


GrayHatter:
Weak commit description is weak...

Is there a parser somewhere? Or is the AST the only part that exists today?

robinlinden:
The AST and execution of the AST are the only parts that exists right now. I have a parser that I'm still working on, but it's not ready for anything yet.

mkiael (approved):

PR #450, opened 2023-01-20
Replace guard with pragma

GrayHatter:
Sample suggestion, not sure if there's other reasons for this guard.


robinlinden:
We only use conventional header guards as `#pragma once` is a non-standard extension. Once we update the code to use C++ modules we won't be needing header guards or `#pragma once`, so this isn't anything we're changing now.

PR #451, opened 2023-01-20
Add backspace for quick navigation during debugging

GrayHatter:


robinlinden (changes requested):
Update the commit message to follow convention, and also to say that you're mapping backspace to the navigate back action. And there's no need to say when you personally use it in the commit message; it's a mapping that all browsers support.

GrayHatter:
will reconsider later

PR #452, opened 2023-01-20
browser/gui: Support disabling writing to disk

GrayHatter:
ImGui by default will write state to a file in the current working
directory. This is undesirable for a number of reasons. i.a. the
behavior of a gui app shouldn't change when called from a different
directory.

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -127,6 +127,12 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
         std::abort();
     }
 
+    if (std::getenv("HST_DISABLE_DISK_IO")) {
+        // TODO(robinlinden): Support for things like HST_DISABLE_DISK_IO=0 to
+        // re-enable IO.
+        ImGui::GetIO().IniFilename = nullptr;
+    }
+
     canvas_->set_viewport_size(window_.getSize().x, window_.getSize().y);
 
     engine_.set_layout_width(window_.getSize().x / scale_);


browser/gui/app.cpp
```diff
@@ -123,6 +123,8 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
         spdlog::critical("imgui-sfml initialization failed");
         std::abort();
     }
+    // ImGui tries to write to the current working directory. This is problematic for a number of reasons.
+    ImGui::GetIO().IniFilename = std::getenv("HASTUR_STATE_FILENAME");
```
robinlinden, line 128: I've thought a bit more about it, and I don't think this is how we want to do this. Not saving anything by default isn't very reasonable, and this env var won't work once we start e.g. caching things and saving browser data outside of whatever imgui is doing right now.

How about either making this respect XDG, and/or adding an env var or flag that tells the browser to never touch the disk? Both seem like reasonable options and things that we'd want in the future as well.

GrayHatter: I have a few worries about imgui's IO design. I'm assuming it's gonna be short term thing, I don't want to allocate to many cycles for it if it's gonna be temporary. I'm happy to make conf write an opt-out envvar if that's better? But I don't want to have to teach it about XDG?

robinlinden: It being an disk IO opt-out env var is fine, but we'll probably use the same env var to block all disk IO, so give it a good name. :P

XDG was just a thought and something someone else can add if they ever feel like it. We'll need to teach hastur a bit about XDG in the future for when we actually need to store browser settings and caches and things either way.

GrayHatter: > XDG was just a thought and something someone else can add if they ever feel like it. We'll need to teach hastur a bit about XDG in the future for when we actually need to store browser settings and caches and things either way.

Agreed, I actually considered if now was the right time to teach hastur about XDG, but the end decision was "soon" over now.

browser/gui/app.cpp
```diff
@@ -123,6 +123,12 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
         spdlog::critical("imgui-sfml initialization failed");
         std::abort();
     }
+    // ImGui tries to write to the current working directory. This can be problematic for a number of reasons,
+    // so it can be prevented via an env var.
+    if (std::getenv("HASTUR_FORBID_DISKIO")) {
```
robinlinden, line 128: `HST_` instead of `HASTUR_` to match the env var (`HST_SCALE`) we already support, and `DISK_IO` since that's not one word.

browser/gui/app.cpp
```diff
@@ -123,6 +123,12 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
         spdlog::critical("imgui-sfml initialization failed");
         std::abort();
     }
+    // ImGui tries to write to the current working directory. This can be problematic for a number of reasons,
```
robinlinden, line 126: This comment doesn't really add anything. ImGui will try to write where we tell it to write, and you don't write why it's problematic. I'd just get rid of it.

browser/gui/app.cpp
```diff
@@ -123,6 +123,12 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
...
+    // so it can be prevented via an env var.
+    if (std::getenv("HASTUR_FORBID_DISKIO")) {
+        // TODO we need a wrapper to support bool & string env vars
```
robinlinden, line 129: Why? `std::getenv` returns a `char *` which you can wrap in a `string_view` without any allocations or anything if you need a string.

robinlinden (changes requested):

browser/gui/app.cpp
GrayHatter: Based on the assumption that eventually we'd like to support a number of env settings where the string may be one of `"0", "1", "2", "true", "FALSE", "T", "false "` etc. and it'd be user hostile to only permit one flavor of the above, or do the wrong thing when someone tries to debug and changes `HST_FORBID_IO_DISK=0" to disable forbidding, or enable IO.

browser/gui/app.cpp
GrayHatter: gone

robinlinden (approved):

PR #454, opened 2023-01-20
ci: Set up gitlint to enforce the commit message format we're using

### LINKED ISSUE #22: Add a CI job checking that the commit message format is followed
The commit messages should be in the format `category: Fix thing`.

E.g.
`build: Fix macOS build`
`protocol: Support Gemini`
`render: Render text`

robinlinden:
Resolves #22

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,7 @@
+[general]
+ignore=body-is-missing
+
+# TODO(robinlinden): Better way of documenting and setting this up.
+# Each commit must start with the main area it affects.
+[title-match-regex]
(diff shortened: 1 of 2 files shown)


.gitlint
```diff
@@ -0,0 +1,7 @@
+[general]
+ignore=body-is-missing
+
+# TODO(robinlinden): Better way of documenting and setting this up.
+# Each commit must start with the main area it affects.
```
GrayHatter, line 5: If only there was a way to ask git for all commits that touch a file/directory

GrayHatter (changes requested):
I'd assume it's better to whitelist the rules that are actually sane.

mkiael (approved):

.gitlint
robinlinden: We're doing it this way because of the parallel implementations of things, along with multiple tokenizers and parsers etc. Instead of "handle comments in css tokenization in the old css library" we write "css: handle comment tokenization". Makes disambiguating between the things shorter and normalized.

PR #455, opened 2023-01-21
Make debug windows right edge sticky

GrayHatter:
Now when the right edge of any debug window in the list is at the edge
of the viewport width, it will stick there through resizes.

There's a bug where if the resize change is enough to "capture" a window
that wasn't intended to be sticky it will capture. Given this didn't
happen much during testing, I don't think the extra state + code
required is worth it for these windows.


browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
     engine_.set_layout_width(windowSize.x / scale_);
 }
 
+void App::sticky_windows(sf::Event::SizeEvent size) {
```
robinlinden, line 160: `const &`. We don't want or need to copy this.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+    const char *names[] = {
```
robinlinden, line 161: std::array, use a range-based for-loop to iterate over the names.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+        "Navigation",
+        "HTTP Response",
+        "DOM",
+        "Stylesheet",
+        "Layout",
+        NULL,
+    };
+    
+    int delta = k_width - size.width;
+    for (uint8_t i = 0;; ++i) {
```
robinlinden, line 171: `std::uint8_t`. There's no guarantee that <cstdint> places things in the global namespace, and once we start using modules, it's guaranteed that this won't be in the global namespace.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+    const char *names[] = {
+        "Navigation",
+        "HTTP Response",
+        "DOM",
+        "Stylesheet",
+        "Layout",
+        NULL,
+    };
+    
+    int delta = k_width - size.width;
```
robinlinden, line 170: This is UB on the first call to this function since k_width isn't initialized.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+    for (uint8_t i = 0;; ++i) {
+        const char *name = names[i];
+        if (!name) {
+            break;
+        }
+        ImGuiWindow *win = ImGui::FindWindowByName(name);
+        if (!win) {
```
robinlinden, line 177: Isn't this a programming error if it ever happens? Until we have a saner way of keeping the names in sync, just `assert()` that we have the window we're looking for.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+            continue;
+        }
+        int edge = (int)win->Pos.x + (int)win->Size.x;
+        if (edge == k_width) {
+            ImGui::SetWindowPos(name, {win->Pos.x - delta, win->Pos.y});
+        }
+    }
+    k_width = size.width;
+    k_height = size.height;
```
robinlinden, line 186: k_height is only ever written to.

browser/gui/app.h
```diff
@@ -74,7 +75,9 @@ class App final {
 
     void navigate();
     void layout();
-
+    void sticky_windows(sf::Event::SizeEvent);
+    int k_width, k_height;
```
robinlinden, line 79: These need better names. k_{width,height} doesn't tell me anything about what they are.

browser/gui/app.cpp
```diff
@@ -156,6 +157,35 @@ void App::set_scale(unsigned scale) {
...
+    for (uint8_t i = 0;; ++i) {
+        const char *name = names[i];
+        if (!name) {
+            break;
+        }
+        ImGuiWindow *win = ImGui::FindWindowByName(name);
+        if (!win) {
+            continue;
+        }
+        int edge = (int)win->Pos.x + (int)win->Size.x;
```
robinlinden, line 180: static_cast instead of C-style cast.

robinlinden (changes requested):

browser/gui/app.cpp
GrayHatter: Aren't integer members 0 if NOS in c++?

browser/gui/app.cpp
GrayHatter: > Isn't this a programming error if it ever happens? 

No, because we might decide that windows are ephemeral. I can think of plenty of reasons we might want to destroy some # of them at runtime.

Assert will crash the app, window stickiness should be a "best effort" feature. No reason to make the app user hostile when it doesn't need to be.

browser/gui/app.cpp
robinlinden: No, unless set in e.g. a constructor they're uninitialized and reading from them is UB.

browser/gui/app.cpp
robinlinden: Right now, all windows always exist. If that changes in the future it will be because a developer changes it and they probably need to have a look at this code and think about what they're doing.

Since you made me look at the imgui internals now, did you have a look at using `GImGui.Windows` for this so we don't have to hard-code a list of window names?

browser/gui/app.cpp
robinlinden: This is still true.

browser/gui/app.cpp
GrayHatter: good enough?

browser/gui/app.cpp
GrayHatter: yes/no?

PR #459, opened 2023-01-30
url: Add url library

Zer0-One:
Adds a blob generation function for generating blob URLs according to
the W3C File API spec: https://w3c.github.io/FileAPI/#blob-url-entry

I will follow this PR up with another one that moves the IP serialization functions to util/string.h .

### DIFF
--- a/url/url.h
+++ b/url/url.h
@@ -1,0 +1,40 @@
+// SPDX-FileCopyrightText: 2023 David Zero <zero-one@zer0-one.net>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#ifndef URL_URL_H_
+#define URL_URL_H_
+
+#include <array>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <variant>
+
+namespace url {
+
+enum class HostType { DnsDomain, Ip4Addr, Ip6Addr, Opaque, Empty };
+
+struct Host {
+    HostType type;
+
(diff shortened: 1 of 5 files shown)


url/url.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <cstdint>
+#include <optional>
+#include <span>
+#include <string>
+#include <variant>
+
+namespace url {
+
+enum class HostType { DNS_DOMAIN, IP4_ADDR, IP6_ADDR, OPAQUE, EMPTY };
```
robinlinden, line 16: `DnsDomain, Ip4Addr, Ip6Addr, Opaque, Empty` to follow name conventions.

url/url.h
```diff
@@ -0,0 +1,40 @@
...
+
+struct Host {
+    HostType type;
+
+    std::variant<std::string, std::uint32_t, std::span<std::uint16_t, 8>> data;
```
robinlinden, line 21: `std::span` -> `std::array`.
A data structure that owns data feels like it makes more sense here since relying on the user to keep the Ipv6-addr alive separately from this struct isn't very intuitive.

robinlinden (changes requested):
Looks good aside from 2 minor nitpicks.

url/url.h
Zer0-One: ah, that's right. whoops

robinlinden (approved):
Looks great!

PR #460, opened 2023-01-30
util/string: Move IP address serialization to util

Zer0-One:


Zer0-One:
It looks like these functions are valid constexpr in clang >= 14 and gcc >= 12, but gcc 10 fails. Do these even really belong in the header, given that they're a bit meatier then most of the other stuff in there?

robinlinden (approved):
Can't be merged before the nuking-of-gcc-10 PR is, but LGTM.

robinlinden:
I thought `constexpr std::string` was in gcc-11, but no such luck. Imo we can drop gcc-11 and target gcc-12-or-higher since Ubuntu 22.04, the latest LTS and the one we support, ships gcc 12.

robinlinden:
Maybe just mark these functions as `inline` for now and add a `TODO(01): constexpr once we drop gcc 11` or something?

Zer0-One:
I specified `inline`, and I'll switch it back over to `constexpr` once we figure out gcc-12.

PR #461, opened 2023-01-30
css: Give the html and body elements the correct display value

robinlinden:


css/default.css
```diff
@@ -1,3 +1,9 @@
+/* https://html.spec.whatwg.org/multipage/rendering.html#the-page */
```
GrayHatter, line 1: Can we clone these locally somewhere/somehow? I hate depending on 3rd party docs that might vanish, or change.

GrayHatter (approved):

css/default.css
Zer0-One: Probably wouldn't be hard to add a CI step to make sure that references to specs are valid/live URLs. But I don't think it's a big deal.

GrayHatter: depending on 3rd party servers to do what you expect is problematic... 

e.g. https://github.com/robinlinden/hastur/pull/462

robinlinden: I'm with zero-one. Checking links for liveliness would be nice, and I don't think mirroring every 3rd-party thing we link to or depend on would be worth the trouble.

mkiael (approved):

PR #462, opened 2023-01-30
Replace http fetch of spdlog dep with git clone

GrayHatter:
Github doesn't promise byte for byte archives of specific commits, so
it's likely better to fetch this dep via git where we can specify the
specific commit we want to use.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,4 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 # Bazel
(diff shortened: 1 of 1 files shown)


robinlinden (commented):
We verify that what we get is byte-for-byte what we expect using sha256.
http_archive is a lot faster than git_repository.
Google depends on GitHub archives using /archive/commit when they need something that's not released, e.g. boringssl.

What's your source for that GitHub doesn't just use git-archive for exports the same way they do for release archives?

GrayHatter:
git is 5% slower and too scary.... per conversation with @robinlinden we'll revisit when github breaks this a 3rd time

robinlinden:
No, per conversation with me
* the Bazel best-practices have always recommended http_archive over git_repository as git_repository doesn't work with the builtin repository cache in Bazel (and it's a lot more than 5% slower on Windows which already has our slowest builds)
* GitHub reverted the change that caused your problem and stated that [more info would follow](https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/), so we don't have good info about what GitHub-provided archives are stable (they have stated that e.g. /archive/refs/tags/\<tag\> would be stable in the past: `https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300`)
* and even if we decide to switch to git_repository we'll have to switch a lot more dependencies as you only fixed the one that you didn't have in your local cache, so if GitHub changed their checksums, any clean build performed after this PR would still be failing.

See also:
* `https://github.com/orgs/community/discussions/46034`
* `https://github.com/orgs/community/discussions/45830`

GrayHatter:
> No, per conversation with me

expecting me to actually *remember* our conversations, way too much work :P 

why do we have bazel managing build deps instead of the system?

Zer0-One:
> why do we have bazel managing build deps instead of the system?

It's probably necessary for reproducible builds. It's also nice to be able to make sure all the dependencies are built with `-march=native -O2` instead of for i386 or whatever ancient architecture most platforms target for compatibility reasons.

PR #464, opened 2023-02-01
browser/gui: Handle CSS special backgrounds

robinlinden:
If the html or body element has a non-transparent background set, that should be used as the canvas background and not just as the element background.

This is currently done in hacky way outside of //render right now as doing correctly, not painting over things outside of the web page, without hacks in //render will require //gfx functionality not in place yet.

Before, html background ending where the html element ends:
Before

After, html background filling the entire canvas:
after

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -6,6 +6,7 @@
 
 #include "css/rule.h"
 #include "dom/dom.h"
+#include "gfx/color.h"
 #include "gfx/opengl_canvas.h"
 #include "gfx/painter.h"
 #include "render/render.h"
@@ -18,6 +19,7 @@
 #include <imgui_stdlib.h>
 #include <spdlog/spdlog.h>
 
+#include <algorithm>
 #include <cmath>
 #include <cstdlib>
 #include <functional>
@@ -25,6 +27,7 @@
 #include <sstream>
 #include <string_view>
 #include <utility>
+#include <variant>
 
 using namespace std::literals;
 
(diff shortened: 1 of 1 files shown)


browser/gui/app.cpp
```diff
@@ -470,9 +473,41 @@ void App::run_layout_widget() const {
...
+    // If html or body has a background set, use that as the canvas background.
+    // TODO(robinlinden): This should be done in //render, but requires new
+    //                    //gfx APIs that I want to think a bit about.
+    if (auto html_bg = engine_.layout().get_property<css::PropertyId::BackgroundColor>();
+            html_bg != gfx::Color::from_css_name("transparent")) {
+        window_.clear(sf::Color(html_bg->as_rgba_u32()));
+        return;
+    }
+
+    auto body = std::ranges::find_if(engine_.layout().children, [](layout::LayoutBox const &c) {
```
GrayHatter, line 494: isn't it time to write a tree selector api?

GrayHatter (approved):
LGTM

browser/gui/app.cpp
robinlinden: I have written `dom::nodes_by_xpath`, and I have extended it to work with both the style and layout tree, but all of that stuff is blocked waiting for reviews. Any improvements to that function will automagically work for the trees wrapping dom nodes once I've finished merging the changes to it.

TL;DR: it was time a while ago, and it is done*. :P

mkiael (approved):

PR #466, opened 2023-02-04
ci: Add a check for link liveness

Zer0-One:
This might be a bit finicky for a bit; will probably require one or two updates in the future as we add different kinds of links in different places and discover that it doesn't match something correctly.

An alternative to this would be to add an explicit tag that I can match on which indicates that a check should be performed. e.g, "[live-check] https://example.com".

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -193,6 +193,12 @@ jobs:
       - run: shfmt -i 2 -w $(shfmt -f .)
       - run: git diff --exit-code
 
+  link-liveness:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: grep --recursive --no-filename --only-matching --exclude=WORKSPACE --exclude=*test.cpp --exclude=ci.yaml 'https://[^)(}{", ]*' | grep -v '^https://$' | xargs wget --spider
+
   gitlint:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -193,6 +193,12 @@ jobs:
       - run: shfmt -i 2 -w $(shfmt -f .)
       - run: git diff --exit-code
 
+  link-liveness:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - run: grep -R -h -o --exclude-dir=bazel-* --exclude=*test.cpp --exclude=ci.yaml 'https://[^)(}{", ]*' | grep -v '^https://$' | xargs wget --spider
```
robinlinden, line 200: Maybe `--recursive --no-filename --only-matching` instead of the shorthand? Makes looking at this and understanding it without docs easier.

You can probably also skip excluding `bazel-*` since those dirs are only created after the first bazel invocation, and maybe add `--exclude=WORKSPACE` since all of those links are checked when we build anyway.

robinlinden (commented):
Great initiative and looks good!

.github/workflows/ci.yaml
Zer0-One: ah yeah, that's a good point

robinlinden (approved):

PR #472, opened 2023-02-08
engine: add security enforcement by default

GrayHatter:
Prevent requesting css files from a different scheme than the host.

This naive implementation does prevent https -> http downgrades, but
also prevents http -> upgrades... Eventually, someone should probably
fix that


Zer0-One:
`uri/` is going to be scrapped once i'm done implementing the new URL parser, so probably best not to put much effort into it. Also, I don't think this makes sense per-se; if you're using a base to complete the URL, then the given URL is a path segment, it won't have a scheme.

GrayHatter:
@Zer0-One sounds fine to me :)

PR #479, opened 2023-02-10
Enable/fix more clang-tidy checks, fix clang-tidy not running on headers

robinlinden:


html2/tokenizer.cpp
```diff
@@ -101,6 +101,7 @@ void Tokenizer::set_state(State state) {
     state_ = state;
 }
 
+// NOLINTNEXTLINE(google-readability-function-size)
```
GrayHatter, line 104: missing rational

GrayHatter (approved):

html2/tokenizer.cpp
robinlinden: Good point. Added my opinion. :P

mkiael (approved):

PR #481, opened 2023-02-11
html: inject html to non-html content

GrayHatter:


GrayHatter:
before 
screenshot

after 
screenshot

html/parser.cpp
```diff
@@ -134,6 +138,15 @@ void Parser::operator()(html2::EndOfFileToken const &) {
     if (!open_elements_.empty()) {
         spdlog::warn("EOF reached with [{}] elements still open", open_elements_.size());
     }
+
+    // TODO this is a hacky way of allowing rendering of raw text without a real
```
robinlinden, line 142: Instead of this, if you don't want to add content-type support, can you hack this in in engine by looking at the file-ending of the url we're loading? It's a hack I already have in my local develop-branch so I can work on image rendering (by looking at urls ending in .png and setting up a nice web page wrapping image).

We shouldn't be passing non-html through the html system, even if it happens to not explode, and relying on it doesn't feel good.

robinlinden (changes requested):

PR #482, opened 2023-02-11
wasm: Add support for decoding unsigned leb128 numbers

robinlinden:
This seems to be the number format they decided to use in WebAssembly. I only added support for decoding it into `uint32_t` for now, but it's trivial to update the decoding to be templated on what kind of type you want the leb128-number decoded into.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -4,4 +4,4 @@ ignore=body-is-missing
 # TODO(robinlinden): Better way of documenting and setting this up.
 # Each commit must start with the main area it affects.
 [title-match-regex]
-regex=^(browser|bzl|css|css2|dom|dom2|engine|etest|geom|gfx|html|html2|img|js|layout|net|os|protocol|render|style|tui|uri|url|util|all|build|ci|deps|doc|meta)(/.*|\+.*)?:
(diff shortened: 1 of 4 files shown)


Zer0-One (approved):

Zer0-One:
LGTM. My only comment is that I'm pretty sure `sizeof(uint8_t)` is guaranteed to be `1`, so maybe it's more readable to just say that.

robinlinden:
> My only comment is that I'm pretty sure sizeof(uint8_t) is guaranteed to be 1, so maybe it's more readable to just say that.

In general I agree with you for one-ofs, but a lot of the rest of `//wasm/...` also uses `is.read(tgt, sizeof(tgt))`, so this keeps things looking roughly the same, and you never have to look at the size of the type to know that the `is.read()`-call is fine.

GrayHatter:
> `sizeof(uint8_t)` is guaranteed to be `1`

[citation needed] because I don't think this is true for all arch... is it?

Zer0-One:
> > `sizeof(uint8_t)` is guaranteed to be `1`
> 
> [citation needed] because I don't think this is true for all arch... is it?

In C it is. I'm not 100% sure about C++, but I imagine it carries over. `uint8_t` is guaranteed to be 8 bits, and `char` is guaranteed to be at least 8 bits. sizeof() calculates size in char-sized units (which means even if a char is 99999 bits, it still amounts to 1 byte according to the compiler), which means `sizeof(uint8_t)` always returns `1`, as long as `uint8_t` is actually defined, and as long as sizeof() doesn't return 0 when something actually has a size > 0.

PR #483, opened 2023-02-11
render css2

GrayHatter:
yes, I know this is ugly AF; only sharing for comments/opinions. If this idea isn't hated, I can write something good that's actually mergable :)


GrayHatter:
updated to drop the commit from the orig-base branch `render-css` see #481

PR #491, opened 2023-02-16
wasm: Implement parsing the export section 

robinlinden:


Zer0-One (approved):
LGTM. Though I'm not totally sure why you use `auto` instead of specifying the exact type (`uint32_t` in this case) for the decoded vector lengths and export indexes. Would this be an issue if you add more specializations for `Uleb128::decode_from()` later?

robinlinden:
> LGTM. Though I'm not totally sure why you use `auto` instead of specifying the exact type (`uint32_t` in this case) for the decoded vector lengths and export indexes. Would this be an issue if `Uleb128::decode_from()` can return different sizes in the future?

Once we add support for other sizes to Leb128, the call-sites will need updating to specify what size they want:
```C
// Before
auto idx = Uleb128::decode_from(is);
// or
std::optional<std::uint32_t> idx = Uleb128::decode_from(is);
```


```C
// After
auto idx = Leb128<std::uint32_t>::decode_from(is);
```

So there's no risk of us accidentally getting the wrong widths in places when we update our leb128-implementation to handle other widths and signed integers.

PR #503, opened 2023-02-23
layout: Support xpath-lookup for LayoutBox

robinlinden:


layout/layout_test.cpp
```diff
@@ -1139,5 +1139,40 @@ int main() {
...
+                        style::StyledNode{.node = div_node,
+                                .children{
+                                        style::StyledNode{.node = p_node},
+                                        style::StyledNode{.node = text_node},
+                                }},
+                },
+        };
+        auto layout = layout::create_layout(styled_node, 123).value();
+
+        // Verify that we a shady anon-box to deal with in here.
```
mkiael, line 1162: Wording

mkiael (approved):

layout/layout_test.cpp
robinlinden: Whoops. I accidentally a word. :( That's what I get for writing comments. Good catch!

PR #506, opened 2023-02-25
Add dependency on icu

Zer0-One:
icu will handle IDNA and misc unicode stuff that requires lugging around giant mapping tables that need to stay up-to-date with the unicode standard.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -100,6 +100,18 @@ http_archive(
 )
 
 http_archive(
+    name = "icu",  # Unicode-DFS-2016
+    build_file = "//third_party:icu.BUILD",
+    patch_cmds = [
+        "rm source/common/BUILD.bazel",
+        "rm source/stubdata/BUILD.bazel",
+    ],
+    sha256 = "a2d2d38217092a7ed56635e34467f92f976b370e20182ad325edea6681a71d68",
+    strip_prefix = "icu",
+    url = "https://github.com/unicode-org/icu/releases/download/release-72-1/icu4c-72_1-src.tgz",
+)
+
+http_archive(
     name = "imgui",  # MIT
     build_file = "//third_party:imgui.BUILD",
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -99,6 +99,18 @@ http_archive(
     url = "https://github.com/nigels-com/glew/releases/download/glew-2.2.0/glew-2.2.0.tgz",
 )
 
+http_archive(
+    name = "icu",  # Unicode Inc
```
robinlinden, line 103: License should be the SPDX identifier.

third_party/icu.BUILD
```diff
@@ -0,0 +1,33 @@
...
+            "-Isource/common/",
+            "-Isource/stubdata/",
+        ],
+        "//conditions:default": [
+            "-frtti",
+            "-Isource/common/",
+            "-Isource/stubdata/",
+        ],
+    }),
+    linkopts = ["-ldl"],
```
robinlinden, line 27: `-ldl` only on Linux.

robinlinden (changes requested):
We're getting ~40k lines of warnings like 
```
In file included from external/icu/source/common/ubidiwrt.cpp:23:
In file included from external/icu/source/common/unicode/ustring.h:22:
external/icu/source/common/unicode/putil.h(29,10): warning: #include resolved using non-portable Microsoft search rules as: external/icu/source/common/unicode/utypes.h [-Wmicrosoft-include]
#include "unicode/utypes.h"
         ^
```
in the clang-cl ci build, so that's something we need to look at before merging this. :P Maybe just an `includes = ["source/common"]` or something that's needed, haven't seen that warning before.

Zer0-One:
`cl : Command line error D8021 : invalid numeric argument '/Wno-microsoft-include'`

I can't figure out how to shut the compiler up about that warning.

Zer0-One:
Warnings fixed

robinlinden (approved):
Looks good!

PR #518, opened 2023-03-03
Dedup URLs for link liveness check

Zer0-One:
This cuts down the number of checked URLs by almost half, since so many are the same page with different anchors.


robinlinden (approved):
Neat. Good idea!

PR #521, opened 2023-03-04
doc: Replace moved links with their new locations

robinlinden:


GrayHatter (approved):
isn't this `docs` instead of `all`?

robinlinden:
> isn't this `docs` instead of `all`?

A bit of both, I guess. I was thinking of docs as README.md, ARCHITECTURE.md, and misc developer documentation, but I think docs makes sense for this too. Will switch. :P

mkiael (approved):

PR #527, opened 2023-03-07
util: Add a helper for generating crc32 checksums

robinlinden:
I need this for some of the parsing bits I'm working on. :P


mkiael:
I think this looks good. Nice to be able to express a size constraint it in the API like that.

mkiael (approved):

PR #530, opened 2023-03-10
deps: Add tl::expected

robinlinden:
We'll replace this with std::expected once our compilers have implemented more of C++23.


mkiael (approved):
Nice!

PR #532, opened 2023-03-11
img: Implement parsing the width and height of GIFs

robinlinden:


img/gif.cpp
```diff
@@ -0,0 +1,104 @@
+// SPDX-FileCopyrightText: 2023 Robin Lindén <dev@robinlinden.eu>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "img/gif.h"
+
+#include <array>
+#include <bit>
```
mkiael, line 8: Nothing used from `<array>` and `<bit>`, right?

mkiael (approved):

img/gif.cpp
robinlinden: Good catch! They're from an earlier implementation when I'd misread the spec. I'd used std::byteswap to flip width and height from LSB to MSB.

PR #533, opened 2023-03-11
engine: Add a test that fetches stylesheets in parallel

robinlinden:
This exercises the multithreading in //engine so tsan can detect any
potential errors.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -61,6 +61,7 @@ build:asan --action_env=LSAN_OPTIONS=report_objects=1
 
 build:tsan --strip=never
 build:tsan --features=tsan
+build:tsan --action_env=TSAN_OPTIONS=halt_on_error=1
 
 build:ubsan --strip=never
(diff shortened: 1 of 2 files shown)


Zer0-One:
Looking at this, I wonder if it wouldn't be a better idea to use non-blocking IO instead of threading. Thoughts?

mkiael (approved):

robinlinden:
> Looking at this, I wonder if it wouldn't be a better idea to use non-blocking IO instead of threading. Thoughts?

Definitely better, and in the backlog. :P We don't really need threads for things like networking or reading from disk, but it's what was the easiest thing to do at the time and didn't require any additional dependencies. There is a [paper](https://wg21.link/p2300) adding support for things like that that might make it into C++26, so when we do something like it it'll probably be based on [libunifex](https://github.com/facebookexperimental/libunifex/) which is a prototype of it.

PR #542, opened 2023-03-16
Enable more clang-tidy checks

robinlinden:


Zer0-One (approved):

Zer0-One:
I'm not too sure about the qualified-auto check; might just be more annoying than anything, but I guess we'll see.

mkiael (approved):

PR #551, opened 2023-03-21
wasm: Add initial instruction parsing

robinlinden:
This is enough to get
```
int test(int* input, int length) {
  int sum = 0;
  for (int i = 0; i < length; ++i) {
    sum += input[i];
  }
  return sum;
}
```
compiled to wasm parsing.

The wat:
```
(module
  (type $type0 (func (param i32 i32) (result i32)))
  (table 0 anyfunc)
  (memory 1)
  (export "memory" memory)
  (export "_Z4testPii" $func0)
  (func $func0 (param $var0 i32) (param $var1 i32) (result i32)
    (local $var2 i32) 
    i32.const 0
    set_local $var2
    block $label0
      get_local $var1
      i32.const 1
      i32.lt_s
      br_if $label0
      loop $label1
        get_local $var0
        i32.load
        get_local $var2
        i32.add
        set_local $var2
        get_local $var0
        i32.const 4
        i32.add
        set_local $var0
        get_local $var1
        i32.const -1
        i32.add
        tee_local $var1
        br_if $label1
      end $label1
    end $label0
    get_local $var2
  )
)
```

### DIFF
--- a/wasm/instructions.cpp
+++ b/wasm/instructions.cpp
@@ -1,0 +1,168 @@
+// SPDX-FileCopyrightText: 2023 Robin Lindén <dev@robinlinden.eu>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "wasm/instructions.h"
+
+#include "wasm/leb128.h"
+
+#include <iomanip>
+#include <iostream>
+#include <istream>
+#include <sstream>
+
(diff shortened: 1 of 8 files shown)


robinlinden:
@mkiael This is ready for review if you're interested in the wasm engine.

PR #552, opened 2023-03-24
deps: Add libunifex

robinlinden:


.github/workflows/ci.yaml
```diff
@@ -33,11 +33,13 @@ jobs:
             version: 14
             bazel: --config asan
 
+          # https://github.com/llvm/llvm-project/issues/49689
           - name: clang-ubsan
             os: ubuntu-22.04
             compiler: clang
-            version: 14
+            version: 15
```
GrayHatter, line 40: not v16?

robinlinden: Only 15 was needed. :P I'll bump all sanitizer-jobs at the same time, unless it's forced, like in this case.

mkiael (approved):

PR #553, opened 2023-03-25
ci: Increase the test timeout when running under valgrind

robinlinden:
This allows us to run everything under valgrind again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,8 +17,7 @@ jobs:
             os: ubuntu-22.04
             compiler: gcc
             version: 12
-            # We exclude @freetype2//:bbox_test as that can take >1 minute to run under valgrind in CI.
-            bazel: --run_under="valgrind --leak-check=full --errors-for-leak-kinds=all --error-exitcode=1 --track-origins=yes --show-leak-kinds=all" -- -@freetype2//:bbox_test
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,8 +17,7 @@ jobs:
             os: ubuntu-22.04
             compiler: gcc
             version: 12
-            # We exclude @freetype2//:bbox_test as that can take >1 minute to run under valgrind in CI.
-            bazel: --run_under="valgrind --leak-check=full --errors-for-leak-kinds=all --error-exitcode=1 --track-origins=yes --show-leak-kinds=all" -- -@freetype2//:bbox_test
+            bazel: --test_timeout=120 --run_under="valgrind --leak-check=full --errors-for-leak-kinds=all --error-exitcode=1 --track-origins=yes --show-leak-kinds=all"
```
GrayHatter, line 20: it's yaml; might as well make this a variable... given it'll probably need a bigger increase soon

GrayHatter (approved):

Zer0-One (approved):

.github/workflows/ci.yaml
robinlinden: This is close to where it's used, and nothing else is anywhere remotely close to hitting the 60s timeout we had before. Leaving it as-is.

PR #554, opened 2023-03-25
[css] Add better error coloring for missing rules

GrayHatter:


robinlinden (commented):
I don't think this is a good idea. Non-deterministic extra code for a very temporary error-case that will go away in favour of whatever other browsers are doing.

PR #557, opened 2023-03-26
ci: Allow code coverage to fluctuate a bit

robinlinden:
This isn't really great, but after we migrated from gcc 10 to gcc 11 for code coverage, gcc has been a bit confused about if e.g. functions' closing braces or even lambda variable declarations are covered when it's literally impossible to not hit them if you call the function.

lambda declaration not covered

closing brace not covered

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,12 @@
 # https://docs.codecov.com/docs/pull-request-comments#disable-comment
 comment: false
+
+coverage:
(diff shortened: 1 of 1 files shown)


robinlinden:
Tested by cherry-picking a branch that has the problem of random end-braces no longer being covered.

Zer0-One (approved):

PR #560, opened 2023-03-28
html2: Add support for rcdata in the tokenizer

robinlinden:


mkiael (approved):
Handling all states now, great work!

PR #579, opened 2023-04-17
browser/gui: Stop re-rendering the GUI when nothing's happening

### LINKED ISSUE #230: Only re-render the GUI when needed
Right now we're rendering at 60 fps regardless of whether we need to or not.

robinlinden:
I would have liked this to be a bool we'd flip when we get an event in
the SFML event loop, but ImGui needs to run for a few iterations every
time it does things.

This reduces idle CPU usage to 0, "busy" CPU usage by a lot, and makes the GUI a lot smoother on displays that can do more than 60 fps.

Resolves #230

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -127,7 +127,6 @@ App::App(std::string browser_title, std::string start_page_hint, bool load_start
     : browser_title_{std::move(browser_title)}, window_{sf::VideoMode(kDefaultResolutionX, kDefaultResolutionY),
                                                         browser_title_},
       url_buf_{std::move(start_page_hint)} {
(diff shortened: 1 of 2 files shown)


Zer0-One:
Do we need to worry about screen tearing in this context? vsync, etc?

robinlinden:
> Do we need to worry about screen tearing in this context? vsync, etc?

Maybe, but I haven't seen any yet, so I didn't want to flip any switched without verifying they're needed. It's just a `window_.setVerticalSyncEnabled(true);` whoever sees tearing will have to add, so no big deal. :P

Zer0-One (approved):

PR #599, opened 2023-05-04
ci: Add a gcc 13 job

robinlinden:


robinlinden:
Neat. libstdc++ was cleaned up to the point where we got compilation errors where we used `uint8_t` instead of `std::uint8_t` (with the `<cstdint>`-include.)

Zer0-One (approved):

PR #601, opened 2023-05-06
doc: Fix broken CSS spec links

robinlinden:
See: https://github.com/w3c/csswg-drafts/issues/8798


GrayHatter (approved):
no one could have predicted this :P

LGTM 

shipit

PR #620, opened 2023-06-03
ci: Downgrade Windows runners to work around broken MSVC setup

robinlinden:
Bazel no longer seems to work in windows-2022 runners as of version 20230517.1, so we'll use windows-2019 as a workaround for now.


robinlinden:
I guess we're using C++ too new for VS2019 now, so never mind. :(

PR #622, opened 2023-06-05
build: Add a Clang 16 code coverage configuration

robinlinden:


robinlinden:
Originally I added a CI job for this as well, but I dropped that for now as it reports most whitespace as missing coverage, even in the middle of functions that otherwise have 100% coverage. Will spend more time looking into it later.

Zer0-One (approved):

PR #623, opened 2023-06-06
ci: Set jobs to only be allowed to read the repo contents

robinlinden:


GrayHatter (approved):
should jobs ever need to write? If not probably should enforce this token side as well?

robinlinden:
> should jobs ever need to write? If not probably should enforce this token side as well?

Not for anything we're doing right now at least. Disabled it in the GitHub GUI as well:
image

mkiael (approved):

PR #638, opened 2023-07-16
meta: Make it possible to run clang-tidy using Bazel

robinlinden:


Zer0-One:
On first glance, this looks like it would run clang-tidy on all third-party deps. Is that right?

robinlinden:
> On first glance, this looks like it would run clang-tidy on all third-party deps. Is that right?

Nah, no way any of our deps would pass our clang-tidy checks, especially since we use clang-tidy to enforce naming conventions that clang-format doesn't handle.

The 3rd-party tests I had to exclude using tags were included because I pulled them into "our targets" using `test_suite` and `alias` in `third_party/BUILD` so that they run together with the rest of our tests.

We can run it on third-party targets if we want, but we'll either have to target them explicitly like `bazel build @freetype2//... --config=clang-tidy` or pull them into the list of targets `//...` in our workspace expands into using aliases or something.

Zer0-One (approved):

PR #641, opened 2023-07-21
url/url: Complete implementation of Origin

Zer0-One:


url/url.h
```diff
@@ -28,15 +28,61 @@ struct Host {
     std::variant<std::string, std::uint32_t, std::array<std::uint16_t, 8>> data;
 
     std::string serialize() const;
+
+    bool operator==(Host const &b) const {
```
robinlinden, line 32: Looks like this can be defaulted?

url/url.h
```diff
@@ -28,15 +28,61 @@ struct Host {
...
     Host host;
     std::optional<std::uint16_t> port;
     std::optional<std::string> domain;
-    // Need this placeholder until I figure out what "opaqueness" means for an origin in this context
-    bool opaque;
+    // If opaque, then this Origin should serialize to (null). All opaque origins are equal to each other, and not equal
+    // to all non-opaque origins.
+    bool opaque = false;
+
+    std::string serialize();
```
robinlinden, line 50: This and `effective_domain` should probably both be `const`?

robinlinden (changes requested):

url/url.h
Zer0-One: I thought so initially too, but it fails to compile.

```
./url/url.h: In member function 'bool url::Origin::operator==(const url::Origin&) const':
./url/url.h:60:44: error: no match for 'operator==' (operand types are 'const url::Host' and 'const url::Host')
   60 |             if (scheme == b.scheme && host == b.host && port == b.port) {
      |                                       ~~~~ ^~ ~~~~~~
      |                                       |         |
      |                                       |         const url::Host
      |                                       const url::Host
```

Zer0-One: nevermind, I didn't know this was a thing: https://en.cppreference.com/w/cpp/language/default_comparisons

url/url.h
```diff
@@ -28,15 +28,55 @@ struct Host {
     std::variant<std::string, std::uint32_t, std::array<std::uint16_t, 8>> data;
 
     std::string serialize() const;
+
+    bool operator==(Host const &b) const = default;
```
robinlinden, line 32: No need to name the argument here.

robinlinden (changes requested):
Like in the last PR, you can replace `url/url:` in the commit message w/ `url:`.

One more test is needed to make CI (and me) happy. :P

robinlinden (approved):

PR #664, opened 2023-08-12
url: Implement WPT url tests

Zer0-One:
Also fixes the bugs that the tests uncovered.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -156,6 +156,14 @@ http_archive(
     url = "https://github.com/eliasdaler/imgui-sfml/archive/v2.6.tar.gz",
 )
 
+http_archive(
+    name = "simdjson",  # Apache-2.0
+    build_file = "//third_party:simdjson.BUILD",
+    sha256 = "13a702536e051db612cdca82bf8585f2c69d9c6fd156ef291b170f13202c1b4c",
+    strip_prefix = "simdjson-3.2.2",
+    url = "https://github.com/simdjson/simdjson/archive/refs/tags/v3.2.2.tar.gz",
+)
+
 # https://github.com/glennrp/libpng
 http_archive(
     name = "libpng",  # Libpng
@@ -232,6 +240,14 @@ http_archive(
     sha256 = "a55555b9e553accbab90b6e15d55a9f1fec47fab12af34c05ba3150b61dd59a5",
     strip_prefix = "Vulkan-Hpp-%s" % VULKAN_TAG,
     url = "https://github.com/KhronosGroup/Vulkan-Hpp/archive/v%s.tar.gz" % VULKAN_TAG,
+)
+
+http_archive(
+    name = "wpt",  # BSD-3-Clause
+    build_file_content = "exports_files([\"url/resources/urltestdata.json\"])",
+    sha256 = "0bec4b1c5bbae0f30b83049d61109186b69ce98cb2af32a0f95eb44be96228c8",
+    strip_prefix = "wpt-merge_pr_41277",
+    url = "https://github.com/web-platform-tests/wpt/archive/refs/tags/merge_pr_41277.tar.gz",
 )
 
 # The freedesktop GitLab goes down too often to be trusted.
--- a/url/url.cpp
+++ b/url/url.cpp
@@ -329,7 +329,7 @@ Origin Url::origin() const {
 }
 
 void UrlParser::validation_error(ValidationError err) const {
-    spdlog::debug("url: InputPos: {}, ParserState: {}, Validation Error: {} {}",
+    spdlog::warn("url: InputPos: {}, ParserState: {}, Validation Error: {} {}",
             current_pos(),
             std::to_underlying(state_),
             std::to_underlying(err),
@@ -571,6 +571,9 @@ void UrlParser::state_scheme() {
         state_ = ParserState::NoScheme;
 
         reset();
+
+        // This can underflow pos_; that's ok, because it's incremented again before it's ever used.
+        back(1);
     } else {
         state_ = ParserState::Failure;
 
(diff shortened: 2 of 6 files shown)


url/url.cpp
```diff
@@ -1460,14 +1464,16 @@ std::optional<std::tuple<std::uint64_t, bool>> UrlParser::parse_ipv4_number(std:
...
 
     auto res = std::from_chars(input.data(), input.data() + input.size(), out, r);
 
+    // This deviation from the spec is necessary, because the spec assumes arbitrary precision
     if (res.ec == std::errc::invalid_argument || res.ec == std::errc::result_out_of_range) {
-        spdlog::info("Invalid ipv4 number");
+        spdlog::info("IPv4 number > 2^64");
 
-        return std::nullopt;
+        return {{out, true}};
```
robinlinden, line 1476: `out` isn't modified on error, so this is returning garbage: https://en.cppreference.com/w/cpp/utility/from_chars#Return_value

I can't imagine the tests would pass if this is ever read. Can you return something something saner?

url/url.cpp
```diff
@@ -1460,14 +1464,16 @@ std::optional<std::tuple<std::uint64_t, bool>> UrlParser::parse_ipv4_number(std:
         }
     }
 
+    // TODO(zero-one): Differ width based on largest integer value supported by platform?
     std::uint64_t out;
 
     auto res = std::from_chars(input.data(), input.data() + input.size(), out, r);
 
+    // This deviation from the spec is necessary, because the spec assumes arbitrary precision
```
robinlinden, line 1472: Arbitrary precision sounds like it should only apply to result_out_of_range? Can this be split so it returns failure for invalid argument?

url/url_test.cpp
```diff
@@ -634,6 +637,105 @@ int main() {
...
+                }
+            }
+
+            // Parse input URL
+            std::optional<url::Url> url = p.parse(std::string{input}, base_test);
+
+            if (!failure) {
+                etest::expect(url.has_value(), "Parsing input URL:(" + std::string{input} + ") failed");
+            } else {
+                etest::require(!url.has_value(),
```
robinlinden, line 688: Maybe replace the require w/ expect in all the boring comparisons so we can get more than one error at a time when running the wpt test suite?

url/url_test.cpp
```diff
@@ -634,6 +637,105 @@ int main() {
...
+
+                    continue;
+                }
+            }
+
+            // Parse input URL
+            std::optional<url::Url> url = p.parse(std::string{input}, base_test);
+
+            if (!failure) {
+                etest::expect(url.has_value(), "Parsing input URL:(" + std::string{input} + ") failed");
```
robinlinden, line 686: You need a continue after this, or we hit UB when the rest of the checks in the test dereference `url`.

url/url_test.cpp
```diff
@@ -634,6 +637,105 @@ int main() {
...
+
+        simdjson::ondemand::array arr = doc.get_array();
+
+        for (auto obj : arr) {
+            // Skip strings, those are just comments
+            if (obj.type() == simdjson::ondemand::json_type::string) {
+                continue;
+            }
+
+            bool failure = false;
```
robinlinden, line 657: Maybe rename this to `should_fail` so it's clearer what it means in this context?

url/url_test.cpp
```diff
@@ -634,6 +637,105 @@ int main() {
...
+
+        simdjson::ondemand::parser parser;
+
+        auto json = simdjson::padded_string::load("external/wpt/url/resources/urltestdata.json");
+
+        simdjson::ondemand::document doc = parser.iterate(json);
+
+        simdjson::ondemand::array arr = doc.get_array();
+
+        for (auto obj : arr) {
```
robinlinden, line 651: `auto const &obj : arr`, or we'll copy every element in the array.

url/url.cpp
```diff
@@ -1460,14 +1464,16 @@ std::optional<std::tuple<std::uint64_t, bool>> UrlParser::parse_ipv4_number(std:
         }
     }
 
+    // TODO(zero-one): Differ width based on largest integer value supported by platform?
```
robinlinden, line 1467: Won't an u32 always be enough for an ipv4-address?

Zer0-One: yes, but some of the parser tests expect an individual number to be parsed correctly while failing the URL as a whole. i.e, this implementation has to deviate from the spec unless we use some kind of bignum library to allow numbers with an indefinite number of digits.

url/url_test.cpp
Zer0-One: I added another check to gate the continue; is there a nicer way to use the expect() to avoid doing the comparison again? Doesn't look like it. Maybe we can make the etest functions return a boolean or something?

url/url_test.cpp
robinlinden: Ah, right. You can leave it like this then. Maybe they've overloaded operator[] to something like lookup-or-create like std::map has.

WORKSPACE
```diff
@@ -157,6 +157,14 @@ http_archive(
 )
 
 # https://github.com/glennrp/libpng
```
robinlinden, line 159: This comment belongs w/ libpng.

WORKSPACE
```diff
@@ -235,6 +243,14 @@ http_archive(
 )
 
 # The freedesktop GitLab goes down too often to be trusted.
```
robinlinden, line 245: This comment belongs w/ the freedesktop libs.

url/url_test.cpp
```diff
@@ -634,6 +637,128 @@ int main() {
...
     });
 
+    etest::test("URL parsing: file url with base", [] {
+        url::UrlParser p;
+
+        std::optional<url::Url> file_base = p.parse("file:///usr/bin/vim");
+
+        etest::require(file_base.has_value());
+
+        etest::expect(p.parse("file:usr/bin/emacs", file_base).has_value());
```
robinlinden, line 647: Can you expect_eq these against what this should actually parse to? I could delete the entirety of the file parsing and these 2 tests would still pass. This is both to have working examples and to make sure I don't break things if I touch the url parsing in the future. :P

robinlinden (changes requested):
I think this is good to merge once this set of comments is addressed. Looks good!

robinlinden (approved):
Great work! First part of the codebase that's tested against something like a proper spec-test! Exciting times. :P

PR #702, opened 2023-09-13
idna: Implement decoding punycode to UTF-8

robinlinden:


robinlinden:
@Zer0-One baby steps towards dropping icu.

Zer0-One (approved):

Zer0-One:
It's so new that it's not documented on cppreference yet, but C23 is supposed to introduce some arithmetic macros that do overflow/wrap checking: https://www.gnu.org/software/gnulib/manual/html_node//stdckdint_002eh.html .

robinlinden:
> It's so new that it's not documented on cppreference yet, but C23 is supposed to introduce some arithmetic macros that do overflow/wrap checking: https://www.gnu.org/software/gnulib/manual/html_node//stdckdint_002eh.html .

Neat! Guess we'll see if that makes it into compilers we use before we end up with something like `util::Checked<int>` to nih it while our tooling catches up.

PR #728, opened 2023-10-14
idna: Generate the mapping table needed for uts46

robinlinden:


robinlinden:
@Zer0-One I hacked this into the url parser to see if it's at least somewhat sane, and the tests that only require the uts46 mapping step to pass seem happy with it.

Zer0-One (approved):
Very cool. One step closer to being able to ditch ICU.

PR #729, opened 2023-10-14
deps: Remove ctre

robinlinden:
We haven't used it since dropping it in favour of std::regex in //uri.


Zer0-One (approved):

Zer0-One:
Maybe update the builds that add `-utf-8` explicitly?

Zer0-One (approved):

robinlinden:
> Maybe update the builds that add `-utf-8` explicitly?

Absolutely. I'd missed that we had that requirement outside of ctre! Thank you!

PR #743, opened 2023-10-23
azm: Stub out a simple AMD64 assembler

robinlinden:
The main use-case for this is to be able to optionally jit-compile wasm instead of interpreting it.

### DIFF
--- a/azm/amd64/assembler.h
+++ b/azm/amd64/assembler.h
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
-#ifndef AZM_ASSEMBLER_H_
-#define AZM_ASSEMBLER_H_
+#ifndef AZM_AMD64_ASSEMBLER_H_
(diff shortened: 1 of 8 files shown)


azm/amd64/assembler.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include <cstdint>
+#include <iostream>
+#include <optional>
+#include <utility>
+#include <vector>
+
+namespace azm::amd64 {
+
+enum class Register {
```
Zer0-One, line 16: Might make sense to call this enum `Register32` or something. That way you can implement a mov overload for, e.g, a 64-bit immediate to register, and you won't have to do a check to see if the destination is valid, because you'll be using a `Register64` there.

robinlinden: Yeah, that's a good point, and the same reason there's an `Imm32` and not an `Imm`. :P Thank you!

Zer0-One (approved):

PR #746, opened 2023-10-26
os: Supporting creating executable memory

robinlinden:
This is needed for some ongoing wasm jit-compilation experiments.


GrayHatter (approved):
No major concerns. I might have said it looks way to easy to lose track of this memory, but you already said basically the same thing to me :sweat_smile:

robinlinden:
> No major concerns. I might have said it looks way to easy to lose track of this memory, but you already said basically the same thing to me 😅

Thanks again for having a look! Will split the lib up like we talked about, so we have some separation of concerns in os.h and don't stick exec-memory-management together with getting the os' font dirs.

PR #750, opened 2023-10-31
browser: Send a more truthful User-Agent header

robinlinden:


Zer0-One:
This one I'm not too sure about. I think it might actually make sense to just mimic the Firefox ESR UA string (for privacy reasons) unless the browser ends up supporting features that don't exist in the other big browsers, or if it's missing support for big features the other ones have.

robinlinden:
Yeah, I agree that this is worse for fingerprint-privacy, even if Hastur's probably pretty fingerprintable right now anyway due to the few headers we send and things like what we don't load and the lack of TCP-pipelining.

Let's hold off on merging something like this for now and make the UA-header settable from e.g. a `.hasturrc` and/or the GUI in the future.

PR #761, opened 2023-11-15
engine: Don't follow redirects an infinite number of times

robinlinden:


Zer0-One (approved):
This looks fine, but a question I have is whether or not it makes a difference if the redirects are in fact a loop, or if there's 10 redirects that take you to a different place.

robinlinden:
The number of redirects in a row (even if you're not in a loop) allowed by Firefox and Chrome appears to be 20, but I think this is fine for now.

Will rename the error as this doesn't detect loops. At first I was detecting loops by dumping things in a std::set and making sure every load had a unique (Response, Url), but I feel like something like that can wait until we have a cache setup since just having a limit like this covers loops as well, just with a bit more loads.

PR #767, opened 2023-11-20
layout: Actually measure the width of text runs

robinlinden:
Previously we were just making a number up (50% of the font-width for the width, 100% for the height.) The making-things-up behaviour is preserved in the `type::NaiveType` `type::IType` implementation as it's useful for tests and things like the tui where we don't care about the measurements.

This currently duplicates some font-search code from `gfx::SfmlCanvas`, but `gfx::SfmlCanvas` should be updated to use the new text/font/typeface system.


robinlinden:
As an example, this is the before and after for http://www.iana.org/domains/reserved
before:
before

after:
after

PR #775, opened 2023-11-27
dom2: Remove stale beginning of a spec-compliant dom implementation

robinlinden:
As sad as it is, I figure it's time to nuke this since I'm not thinking of putting any work into this and I think effort put into it right now would be wasted.

This hasn't really been touched since it was stubbed out in 2021 except for clang-tidy warnings being fixed as they're enabled.

I don't think I'd implement this the same way if I were to start it today, but even if we were to decide we love this old design, it's just one revert away.


Zer0-One (approved):
RIP in peace

PR #779, opened 2023-11-30
browser/gui: Prime the font system w/ a 'monospace' font

robinlinden:
DejaVu Sans Mono for Linux for now because that's what happened to live on the machine I ran `fc-match monospace` on. This will be looked up in a more reasonable fashion in the future.

```
robin@q:~/code/hastur$ fc-match monospace
DejaVuSansMono.ttf: "DejaVu Sans Mono" "Book"
```


robinlinden:
Yeah, for Linux we won't be hardcoding fonts like this in the future and probably actually use something like fontconfig (or parse the files) ourselves, and as you say, we'll probably want to care about more things than just a font being monospace.

If `DejaVuSansMono.ttf` isn't found, the font-engine will still fall back to trying to find a font w/ `monospace` in the file name right now. For Windows it might make sense to hardcode a font and let the user override it from the UI somewhere, but the real reason for this is that I need a `monospace` font so I can check my layout work. Feel free to add a monospace font that works well on your system to the Linux side of the array. :P

PR #787, opened 2023-12-10
ci: Add macos build

### LINKED ISSUE #21: Add macOS CI

Zer0-One:
Resolves #21

### DIFF
--- a/third_party/sfml.BUILD
+++ b/third_party/sfml.BUILD
@@ -1,4 +1,4 @@
-load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_cc//cc:defs.bzl", "cc_library", "objc_library")
 
 SFML_DEFINES = [
     "SFML_STATIC",
@@ -10,6 +10,10 @@ cc_library(
     name = "system",
     srcs = glob(["src/SFML/System/*.cpp"]) + select({
         "@platforms//os:linux": glob([
+            "src/SFML/System/Unix/**/*.cpp",
+            "src/SFML/System/Unix/**/*.hpp",
+        ]),
+        "@platforms//os:macos": glob([
             "src/SFML/System/Unix/**/*.cpp",
             "src/SFML/System/Unix/**/*.hpp",
         ]),
@@ -26,6 +30,9 @@ cc_library(
     defines = SFML_DEFINES,
     linkopts = select({
         "@platforms//os:linux": [
+            "-pthread",
+        ],
+        "@platforms//os:macos": [
             "-pthread",
         ],
         "@platforms//os:windows": [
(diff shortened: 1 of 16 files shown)


Zer0-One:
It passes, but it's not building everything because of the targets that are only linux-compatible. e.g `udev-zero` dep for `@sfml//:window` on macos doesn't make any sense, but it gets the build working.

Do you wanna merge this and go from here? This will probably require code changes to get the whole thing building.

third_party/unifex.BUILD
```diff
@@ -35,6 +37,11 @@ cc_library(
     name = src[:-4],
     size = "small",
     srcs = [src],
+    target_compatible_with = select({
+        "@platforms//os:linux": [],
+        "@platforms//os:windows": [],
+        "//conditions:default": ["@platforms//:incompatible"],
```
robinlinden, line 43: Probably only mark this as incompatible with macos, and a comment about why? Looks like they have a macos-build in CI, so I was expecting it to work, but I haven't looked into it at all.

third_party/sfml.BUILD
```diff
@@ -54,6 +61,10 @@ cc_library(
             "src/SFML/Window/Unix/*.cpp",
             "src/SFML/Window/Unix/*.hpp",
         ]),
+        "@platforms//os:macos": glob([
+            "src/SFML/Window/Unix/*.cpp",
+            "src/SFML/Window/Unix/*.hpp",
+        ]),
```
robinlinden, line 67: SFML has macOS-specific implementations of most things: https://github.com/SFML/SFML/tree/ea42079da9b72f81d49bfea476e7c8cd3fff97a0/src/SFML/Window/macOS

If you don't want to deal with that right now, feel free to drop the macos-fixes here and just mark the library as incompatible together with a TODO for someone to deal with later.

Just having most of our things building on macOS in CI is a massive improvement, so I'm more than happy to push some things like this to the future.

robinlinden (changes requested):
Great work! More than one person has looked at macOS-support before, so this is fantastic progress. :)

third_party/sfml.BUILD
Zer0-One: A lot of the macOS-specific code is written in objective-C or objective-C++, and it doesn't look like you can include those sources in a `cc_library` rule. Probably have to use `objc_library` instead, so I'm not sure what the nicest way to do this is.

robinlinden: I *think* a objc_library that sfml:window links when platforms/os:macos would work, but tbh I've only touched that part of Bazel when dealing with Android. Probably just mark it as incompatible for now since that could turn into a pretty large chunk of work if you're unlucky.

Zer0-One:
Now there's only 11 tests that are skipped.

robinlinden (approved):
Super neat! That solves the oldest issue we have. :P

PR #788, opened 2023-12-11
build: Update to Bazel 7

robinlinden:


robinlinden:
Bazel 7 switches gcov-8 and newer to use gcc's branch coverage. It seems a bit broken (how can something like `std::stringstream ss;` only have partial coverage?), but I think we should just bite the bullet for now and see if Clang does any better.

image

Zer0-One (approved):

PR #792, opened 2023-12-13
ci: Add pre-commit checks

### LINKED ISSUE #228: Check that every file ends with a newline character

Zer0-One:
Fixes #228

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
+    hooks:
+      - id: trailing-whitespace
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
```
robinlinden, line 3: Looks like v4.5.0 is the latest version. Is there a reason we're on an older one?

robinlinden (approved):

.pre-commit-config.yaml
Zer0-One: I tried a newer version and it failed, but I didn't try 4.5.0. Let's see.

PR #799, opened 2023-12-17
url: Allow overriding the error-handling

robinlinden:
This enables us to e.g. display parse-errors in the UI and other things
that aren't just printing all errors to the console.

The default behaviour is still what it was before this change, but after this change, we could probably drop the spdlog-dependency of //url and instead set up spdlog-logging in the tests and the browser instead if we want.


Zer0-One (approved):
Wanted to replace that with a callback at some point, but I forgot <_<; the logging was really just there for debugging while I was working on it.

If you wanna drop the spdlog dependency now instead of later, feel free.

robinlinden:
> Wanted to replace that with a callback at some point, but I forgot <_<; the logging was really just there for debugging while I was working on it.

Heh, I had a feeling that was the case when one of the spdlog prints was something like `spdlog::info("IPv6 parse failure 2")`. :P I think keeping all the descriptions and making it possible to use the logging as a consumer of `//url` is neat and useful, but let's not default to it then.

> If you wanna drop the spdlog dependency now instead of later, feel free.

I had a look at dropping the spdlog-dependency of `//url`. PTAL; the codecov patch failure is because we can't test uts46-creation failure, which I think should be impossible (outside of oom) now that we link the data statically.

Zer0-One (approved):

PR #809, opened 2023-12-25
deps/icu: Remove dependency on system icu-config

Zer0-One:
Can probably make this look a little bit neater by putting the build options into a starlark dictionary and then wrapping the genrule in a list comprehension or w/e, but not gonna bother unless you really want that.

### DIFF
--- a/third_party/icu.BUILD
+++ b/third_party/icu.BUILD
@@ -1,4 +1,5 @@
 load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+load("//bzl:defs.bzl", "var_providing_rule")
 
 cc_library(
     name = "common",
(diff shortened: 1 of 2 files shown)


third_party/icu.BUILD
```diff
@@ -172,6 +172,56 @@ cc_binary(
     ],
 )
 
+genrule(
+    name = "pkgdata_inc",
+    outs = ["pkgdata.inc"],
+    cmd = select({
+        "@platforms//os:macos": r"""
+echo "
```
robinlinden, line 180: Maybe move these scripts into something like `PKGDATA_INC_MACOS_CMD = ...`-variables above to make this easier to read? It's not super obvious where one commands ends and the next one starts.

robinlinden (changes requested):
With this change, I guess we're able to drop the `brew install icu`-thing in the macOS CI job?

third_party/icu.BUILD
```diff
@@ -172,6 +172,56 @@ cc_binary(
...
+GENCCODE_ASSEMBLY_TYPE=-a gcc-darwin
+SO=dylib
+SOBJ=dylib
+A=a
+LIBPREFIX=lib
+LIB_EXT_ORDER=.73.2.dylib
```
robinlinden, line 186: 74.1?

third_party/icu.BUILD
```diff
@@ -172,6 +172,56 @@ cc_binary(
...
+COMPILE=clang -DU_ATTRIBUTE_DEPRECATED=   -DU_HAVE_STRTOD_L=1 -DU_HAVE_XLOCALE_H=1 -DU_HAVE_STRING_VIEW=1  -O2 -std=c11 -Wall -pedantic -Wshadow -Wpointer-arith -Wmissing-prototypes -Wwrite-strings  -Qunused-arguments -Wno-parentheses-equality -fno-common -c
+LIBFLAGS=-I/opt/homebrew/Cellar/icu4c/73.2/include  -dynamic
```
robinlinden, line 188: What's this used for? Should it point into the Bazel build tree? Maybe add a comment about what's going on with this pkgdata.inc thing?

third_party/icu.BUILD
```diff
@@ -172,6 +172,56 @@ cc_binary(
...
+GENLIB=clang -dynamiclib -dynamic -O2 -std=c11 -Wall -pedantic -Wshadow -Wpointer-arith -Wmissing-prototypes -Wwrite-strings  -Qunused-arguments -Wno-parentheses-equality
+LDICUDTFLAGS=
+LD_SONAME=-Wl,-compatibility_version -Wl,73 -Wl,-current_version -Wl,73.2 -install_name
```
robinlinden, line 191: More 74?

third_party/icu.BUILD
```diff
@@ -172,6 +172,56 @@ cc_binary(
...
+        "//conditions:default": r"""
+echo "
+GENCCODE_ASSEMBLY_TYPE=-a gcc
+SO=so
+SOBJ=so
+A=a
+LIBPREFIX=lib
+LIB_EXT_ORDER=.74.1
+COMPILE=gcc -D_REENTRANT  -DU_HAVE_ELF_H=1 -DU_HAVE_STRTOD_L=1 -DU_HAVE_XLOCALE_H=0  -DU_DISABLE_RENAMING=1 -DU_ATTRIBUTE_DEPRECATED= -march=native -O2 -pipe -std=c11 -Wall -pedantic -Wshadow -Wpointer-arith -Wmissing-prototypes -Wwrite-strings   -c
+LIBFLAGS=-I/usr/include -DPIC -fPIC
```
robinlinden, line 209: Same thing about the Bazel build tree.

PR #811, opened 2023-12-26
url: Perform RFC3986 normalization on serialization

### LINKED ISSUE #238: Normalize URIs

Zer0-One:
Closes #238


robinlinden (approved):
Looks good, but can you either move the new `util/string.h`-functions into `url.cpp` or add tests for them in `util/string_test.cpp`?

PR #813, opened 2023-12-29
wasm: Add instruction serialization with to_string()

Zer0-One:
idk how much uncommitted work you have on this locally, but I figured I'd start hacking away at it. Mainly curious to know what you think about adding that `kMnemonic` constant to the instructions for when we eventually deal with text-format modules.

### DIFF
--- a/wasm/instructions.cpp
+++ b/wasm/instructions.cpp
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2023 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2024 David Zero <zero-one@zer0-one.net>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
@@ -130,6 +131,9 @@ std::optional<std::vector<Instruction>> parse(std::istream &is) {
                 break;
             case I32Add::kOpcode:
                 instructions.emplace_back(I32Add{});
+                break;
+            case I32Sub::kOpcode:
+                instructions.emplace_back(I32Sub{});
                 break;
             case LocalGet::kOpcode: {
                 auto value = wasm::Leb128<std::uint32_t>::decode_from(is);
--- a/wasm/instructions.h
+++ b/wasm/instructions.h
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2023 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2024 David Zero <zero-one@zer0-one.net>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
@@ -10,6 +11,7 @@
 #include <cstdint>
 #include <iosfwd>
 #include <optional>
+#include <string_view>
 #include <variant>
 #include <vector>
 
@@ -30,22 +32,28 @@ struct MemArg {
 
     std::uint32_t align{};
     std::uint32_t offset{};
+
     [[nodiscard]] bool operator==(MemArg const &) const = default;
 };
 
+// Control instructions
 struct Block;
 struct Loop;
 struct BreakIf;
 struct Return;
 
+// Numeric instructions
 struct I32Const;
 struct I32LessThanSigned;
 struct I32Add;
+struct I32Sub;
 
+// Variable instructions
 struct LocalGet;
 struct LocalSet;
 struct LocalTee;
 
+// Memory instructions
 struct I32Load;
 
 using Instruction = std::variant<Block,
@@ -55,6 +63,7 @@ using Instruction = std::variant<Block,
         I32Const,
         I32LessThanSigned,
         I32Add,
+        I32Sub,
         LocalGet,
         LocalSet,
         LocalTee,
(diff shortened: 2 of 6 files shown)


robinlinden:
The wasm-bits I've mostly been working on recently are the compilation (hence `//azm`) and interpreting, so adding more to the parsing won't clash at all, and there's also the validation step we're supposed to do before any execution that I haven't looked at yet.

wasm/instructions.h
```diff
@@ -108,6 +117,12 @@ struct I32Add {
     [[nodiscard]] bool operator==(I32Add const &) const = default;
 };
 
+struct I32Sub {
+    static constexpr std::uint8_t kOpcode = 0x6b;
+    static constexpr std::string_view kMnemonic = "i32.sub"sv;
```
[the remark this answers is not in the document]
Zer0-One: i like the idea of just having a `to_string()` better.

wasm/instructions.cpp
```diff
@@ -171,4 +174,22 @@ std::optional<std::vector<Instruction>> parse(std::istream &is) {
     }
 }
 
+std::optional<std::string> to_string(Instruction const &inst) {
```
Zer0-One, line 177: How about something like this, where `to_string()` is called recursively to serialize the whole sub-tree? So if you pass in a `Block`, it gives you the whole block, but passing in a `I32Add` gives you just the one `(i32.add)`.

robinlinden: Yeah, that sounds great!

wasm/instructions.cpp
```diff
@@ -171,4 +176,71 @@ std::optional<std::vector<Instruction>> parse(std::istream &is) {
...
+    std::string out;
+
+    if (std::holds_alternative<Block>(inst)) {
```
robinlinden, line 182: If we replace this `if-else if`-chain w/ `std::visit`, we can get a compilation error if any of the variants in `Instruction` are unhandled, and that would allow us to always return a `std::string` instead of an optional one.

wasm/instructions.h
```diff
@@ -23,29 +26,72 @@ struct BlockType {
     };
     std::variant<Empty, ValueType, TypeIdx> value;
     [[nodiscard]] bool operator==(BlockType const &) const = default;
+
+    constexpr std::string to_string() {
```
robinlinden, line 30: Any particular reason you made this a member instead of a free function? It might be nice to maintain the same `constexpr std::string to_string(T const &)` signature that the stdlib uses for all to_string-things?

Zer0-One: I thought maybe it seemed cleaner, but I think I like this style just as well.

robinlinden (changes requested):

Zer0-One:
I added the serialization tests as another suite in main(), but I can split them out into separate functions if you think that would be neater.

wasm/instructions_test.cpp
```diff
@@ -26,7 +27,8 @@ std::optional<InsnVec> parse(std::string s) {
 // https://webassembly.github.io/spec/core/binary/instructions.html
 int main() {
     // NOLINTBEGIN(modernize-raw-string-literal): This is hex data, not 'A'.
-    etest::Suite s{"wasm::instructions"};
+    etest::Suite s{"wasm::instructions parsing"};
+    etest::Suite t{"wasm::instructions serialization"};
```
robinlinden, line 31: I think it'd be nicer if we move this to a separate file, or at least a separate function like you mentioned. Having the parsing and serialization tests interspersed like this is a bit hard to read, and a bonus is that your new tests don't need the clang-tidy nolint applied to all the parsing tests.

I think I'd prefer a separate file since they're very different functionalities, but it's up to you.

wasm/instructions.h
```diff
@@ -25,27 +29,75 @@ struct BlockType {
...
+    std::string out;
+
+    // If offset == 0, omit offset
+    if (ma.offset != 0) {
+        out += "offset=";
+        out += std::to_string(ma.offset);
+    }
+
+    // Natural alignment, omit "align=" phrase
+    if (natural_alignment.has_value() && natural_alignment == ma.align) {
```
robinlinden, line 69: Empty optionals always compare false with values, so you don't need to check has_value before comparing.

wasm/instructions.cpp
```diff
@@ -171,4 +176,101 @@ std::optional<std::vector<Instruction>> parse(std::istream &is) {
...
+
+    void operator()(Loop const &t) {
+        out += Loop::kMnemonic;
+        out += " ";
+        out += to_string(t.type);
+        out += " ";
+
+        for (Instruction const &i : t.instructions) {
+            out += "\n";
+            out += "\t";
```
robinlinden, line 208: This will have the wrong indentation if we have nested loops or blocks. Maybe we can pass the current indentation along in the to_string-calls, or add a member variable to the InstructionStringifyVisitor keeping track of it?

robinlinden (changes requested):

Zer0-One:
I split out the string stuff into its own unit.

robinlinden (approved):

PR #873, opened 2024-02-25
ci: Check that all GitHub links are permalinks

robinlinden:
@Zer0-One do the icu permalinks point to the right things? I went back a few commits in the history of that file, and it looks like they were pointing to the start of the enum (but I don't understand why) and the pkgdata.inc-comment most of the time.


Zer0-One:
yeah, they're good.

Zer0-One (approved):

PR #876, opened 2024-02-28
wasm: Add module validation

Zer0-One:
This validation code assumes we're doing a separate validation pass, and not validating as we parse.

### DIFF
--- a/util/variant.h
+++ b/util/variant.h
@@ -11,6 +11,11 @@
 
 namespace util {
 
+template<typename... T, typename... Ts>
+constexpr bool holds_any_of(std::variant<Ts...> const &v) noexcept {
+    return (std::holds_alternative<T>(v) || ...);
+}
+
 template<typename T, typename... Ts>
 struct Sequence {
     template<typename... VariantTypesT, std::size_t SpanSizeT>
--- a/wasm/validation.cpp
+++ b/wasm/validation.cpp
@@ -1,0 +1,593 @@
+// SPDX-FileCopyrightText: 2024 David Zero <zero-one@zer0-one.net>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "wasm/validation.h"
+
+#include "util/variant.h"
+#include "wasm/instructions.h"
+#include "wasm/types.h"
+#include "wasm/wasm.h"
+
+#include <tl/expected.hpp>
+
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <string_view>
+#include <utility>
+#include <variant>
+#include <vector>
+
+namespace wasm::validation {
+
+using namespace wasm::instructions;
+
+using Unknown = std::monostate;
+using ValueOrUnknown = std::variant<Unknown, ValueType>;
+
+std::string_view to_string(ValidationError err) {
+    switch (err) {
+        case ValidationError::BlockTypeInvalid:
+            return "BlockType of a block or loop is invalid; the type section is undefined, or the type index was "
+                   "out-of-bounds.";
+        case ValidationError::CodeSectionUndefined:
+            return "A code section is required, but was not defined";
+        case ValidationError::ControlStackEmpty:
+            return "Attempted to pop from the control stack, but the control stack is empty";
+        case ValidationError::FuncTypeInvalid:
+            return "Function section references a non-existent type";
+        case ValidationError::FunctionSectionUndefined:
+            return "A function section is required, but was not defined";
+        case ValidationError::FuncUndefinedCode:
+            return "Function body is undefined/missing";
+        case ValidationError::LabelInvalid:
+            return "Attempted to branch to a label which isn't valid";
+        case ValidationError::LocalUndefined:
+            return "Attempted to index a local which isn't defined in the current code entry";
+        case ValidationError::MemoryBadAlignment:
+            return "Attempted a load or store with a bad alignment value";
+        case ValidationError::MemoryEmpty:
+            return "Attempted a load, but memory is empty";
+        case ValidationError::MemorySectionUndefined:
+            return "Attempted a load or store, but no memory section was defined";
+        case ValidationError::TypeSectionUndefined:
+            return "A type section is required, but was not defined";
+        case ValidationError::UnknownInstruction:
+            return "Unknown instruction encountered";
+        case ValidationError::ValueStackHeightMismatch:
+            return "Value stack height on exiting a control frame does not match the height on entry";
+        case ValidationError::ValueStackUnderflow:
+            return "Attempted to pop from the value stack, but stack height would underflow";
+        case ValidationError::ValueStackUnexpected:
+            return "Attempted to pop an expected value from the value stack, but got a different value";
+    }
+
+    return "Unknown error";
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#limits
+constexpr bool is_valid(Limits const &l, std::uint64_t k) {
+    if (l.min > k) {
+        return false;
+    }
+
+    if (l.max.has_value()) {
+        if (l.max > k || l.max < l.min) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#block-types
+constexpr bool is_valid(wasm::instructions::BlockType const &bt, Module const &m) {
+    if (auto const *t = std::get_if<TypeIdx>(&bt.value)) {
+        if (!m.type_section.has_value()) {
+            return false;
+        }
+
+        if (*t >= m.type_section->types.size()) {
+            // Index outside bounds of defined types
+            return false;
+        }
+    }
+
+    return true;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#table-types
+constexpr bool is_valid(TableType const &t) {
+    return is_valid(t.limits, (1ULL << 32) - 1);
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#memory-types
+constexpr bool is_valid(MemType const &mt) {
+    return is_valid(mt, 1ULL << 16);
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#match-limits
+// https://webassembly.github.io/spec/core/valid/types.html#memories
+constexpr bool is_match(Limits const &l1, Limits const &l2) {
+    if (l1.min >= l2.min) {
+        if (!l2.max.has_value()) {
+            return true;
+        }
+
+        if (l1.max.has_value() && l1.max <= l2.max) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#functions
+constexpr bool is_match(FunctionType const &f1, FunctionType const &f2) {
+    return f1 == f2;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#tables
+constexpr bool is_match(TableType const &t1, TableType const &t2) {
+    return is_match(t1.limits, t2.limits) && t1.element_type == t2.element_type;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#globals
+constexpr bool is_match(GlobalType const &g1, GlobalType const &g2) {
+    return g1 == g2;
+}
+
+// https://webassembly.github.io/spec/core/appendix/algorithm.html#validation-algorithm
+struct ControlFrame {
+    Instruction i;
+    std::vector<ValueType> params;
+    std::vector<ValueType> results;
+    std::size_t stack_height = 0;
+    bool unreachable = false;
+};
+
+struct InstValidator {
+    std::vector<ValueOrUnknown> value_stack; // operand stack
+    std::vector<ControlFrame> control_stack;
+
+    void push_val(ValueOrUnknown const &);
+    tl::expected<ValueOrUnknown, ValidationError> pop_val();
+    tl::expected<ValueOrUnknown, ValidationError> pop_val_expect(ValueOrUnknown const &);
+    void push_vals(std::vector<ValueType> const &);
+    tl::expected<std::vector<ValueOrUnknown>, ValidationError> pop_vals(std::vector<ValueType> const &);
+    void push_ctrl(Instruction, std::vector<ValueType>, std::vector<ValueType>);
+    tl::expected<ControlFrame, ValidationError> pop_ctrl();
+    std::vector<ValueType> const &label_types(ControlFrame const &);
+    void mark_unreachable();
+};
+
+void InstValidator::push_val(ValueOrUnknown const &val) {
+    value_stack.push_back(val);
+}
+
+tl::expected<ValueOrUnknown, ValidationError> InstValidator::pop_val() {
+    assert(!control_stack.empty());
+
+    if (value_stack.size() == control_stack.back().stack_height && control_stack.back().unreachable) {
+        return Unknown{};
+    }
+
+    if (value_stack.size() == control_stack.back().stack_height) {
+        return tl::unexpected{ValidationError::ValueStackUnderflow};
+    }
+
+    assert(!value_stack.empty());
+
+    auto val = value_stack.back();
+
+    value_stack.pop_back();
+
+    return val;
+}
+
+tl::expected<ValueOrUnknown, ValidationError> InstValidator::pop_val_expect(ValueOrUnknown const &expected) {
+    auto actual = pop_val();
+
+    if (!actual.has_value()) {
+        return tl::unexpected{actual.error()};
+    }
+
+    if (actual != expected && !std::holds_alternative<Unknown>(*actual) && !std::holds_alternative<Unknown>(expected)) {
+        return tl::unexpected{ValidationError::ValueStackUnexpected};
+    }
+
+    return actual;
+}
+
+void InstValidator::push_vals(std::vector<ValueType> const &vals) {
+    value_stack.reserve(value_stack.size() + vals.size());
+
+    for (ValueType const &val : vals) {
+        value_stack.push_back(val);
+    }
+}
+
+tl::expected<std::vector<ValueOrUnknown>, ValidationError> InstValidator::pop_vals(std::vector<ValueType> const &vals) {
+    std::vector<ValueOrUnknown> popped;
+
+    // TODO(dzero): switch to std::ranges::reverse_view once we drop older toolchains
+    // NOLINTNEXTLINE(modernize-loop-convert)
+    for (auto it = vals.rbegin(); it != vals.rend(); it++) {
+        auto maybe_val = pop_val_expect(*it);
+
+        if (!maybe_val.has_value()) {
+            return tl::unexpected{maybe_val.error()};
+        }
+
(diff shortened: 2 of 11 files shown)


wasm/validation.h
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    return true;
+}
+
+constexpr bool is_valid(wasm::instructions::BlockType &bt, Module const &m) {
+    assert(m.type_section.has_value());
+
+    if (auto *t = std::get_if<TypeIdx>(&bt.value)) {
+        if (m.type_section.value().types.size() <= *t) {
```
robinlinden, line 36: When you know an optional (m.type_section) has a value, prefer unchecked access like `*` or `->`. `.value()` throws an exception or terminates when the optional is empty, whereas the unchecked ones cause UB.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <map>
+#include <string_view>
+#include <variant>
+#include <vector>
+
+namespace wasm::validation {
+
+using namespace wasm;
```
robinlinden, line 22: This doesn't do anything. When looking up symbols in `::wasm::validation`, the compiler will search in `::wasm::validation`, `::wasm`, `::`, in that order.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+using namespace wasm::instructions;
+
+using ValueOrUnknown = std::variant<std::monostate, ValueType>;
+
+// NOLINTNEXTLINE(cert-err58-cpp)
+std::map<ValidationError, std::string_view> const validation_error_str = {
```
robinlinden, line 28: Maybe replace this with a switch where each `ValidationError` case returns a string_view in `to_string` instead? No need for a nolint and no heap allocations needed with that method.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+std::string_view to_string(ValidationError err) {
+    return validation_error_str.at(err);
+}
+
+struct ControlFrame {
+    Instruction i;
+    std::vector<ValueType> params;
+    std::vector<ValueType> results;
+    std::size_t stack_height;
```
robinlinden, line 54: Initialize the fundamental types here so that's not something we need to think about when using the struct.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+    bool unreachable;
+};
+
+namespace {
+
+void push_val(std::vector<ValueOrUnknown> &value_stack, ValueOrUnknown const &val) {
+    value_stack.push_back(val);
```
robinlinden, line 61: I think it might be nicer to have something like:
```c++
struct Validator {
  std::vector<ValueOrUnknown> value_stack;
  std::vector<ControlFrame> control_stack;

  tl::expected<ValueOrUnknown, ValidationError> pop_val();
  ...
};
```

Having to pass these stacks to most functions feels like unnecessary boilerplate. What do you think?

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+tl::expected<ValueOrUnknown, ValidationError> pop_val(
+        std::vector<ValueOrUnknown> &value_stack, std::vector<ControlFrame> &control_stack) {
+    // This check assumes that there's always a frame on the control stack.
```
robinlinden, line 66: `assert(!control_stack.empty())` instead of a comment?

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+    return val;
+}
+
+tl::expected<ValueOrUnknown, ValidationError> pop_val_expect(std::vector<ValueOrUnknown> &value_stack,
+        std::vector<ControlFrame> &control_stack,
+        ValueOrUnknown const &expected) {
+    auto actual = pop_val(value_stack, control_stack);
+
+    // propagate pop_val error upward
```
robinlinden, line 87: No need for comments like this. Error propagation with std::expected isn't unexpected or anything.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (!actual.has_value()) {
+        return tl::unexpected{actual.error()};
+    }
+
+    if (actual != expected && !std::holds_alternative<std::monostate>(*actual)
```
robinlinden, line 92: Maybe add a `using Unknown = std::monospace` alias, and an `is_unknown(ValueOrUnknown const &)`-helper to make the semantics here more clear?

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+            && !std::holds_alternative<std::monostate>(expected)) {
+        return tl::unexpected{ValidationError::ValueStackUnexpected};
+    }
+
+    return actual;
+}
+
+void push_vals(std::vector<ValueOrUnknown> &value_stack, std::vector<ValueType> const &vals) {
+    for (ValueType const &val : vals) {
+        value_stack.push_back(val);
```
robinlinden, line 102: It's good practice to reserve space in containers before doing things like push_back on them in a loop in order to prevent bonus reallocations.

`value_stack.reserve(value_stack.size() + vals.size());` or something means this function will make at most one allocation.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (value_stack.size() != frame.stack_height) {
+        return tl::unexpected{ValidationError::ValueStackHeightMismatch};
+    }
+
+    control_stack.pop_back();
+
+    return frame;
+}
+
+std::vector<ValueType> label_types(ControlFrame &frame) {
```
robinlinden, line 161: return a `const &` to get rid of bonus allocations.

wasm/validation.cpp
```diff
@@ -0,0 +1,373 @@
...
+            auto maybe_vals = pop_vals(value_stack, control_stack, {ValueType::Int32, ValueType::Int32});
+
+            if (!maybe_vals.has_value()) {
+                return tl::unexpected{maybe_vals.error()};
+            }
+
+            push_val(value_stack, ValueType::Int32);
+        }
+        // https://webassembly.github.io/spec/core/valid/instructions.html#variable-instructions
+        else if (std::holds_alternative<LocalGet>(i)) {
```
robinlinden, line 270: `get_if` instead of `holds_alternative` + `get`.

wasm/validation.h
```diff
@@ -0,0 +1,115 @@
...
+
+#include <cstdint>
+#include <string_view>
+#include <variant>
+#include <vector>
+
+namespace wasm::validation {
+
+// https://webassembly.github.io/spec/core/valid/types.html#limits
+constexpr bool is_valid(Limits const &l, std::uint64_t k) {
```
robinlinden, line 22: If these are only used internally in the validation, then consider moving them into the .cpp (they can still be constexpr etc.)

wasm/validation.h
```diff
@@ -0,0 +1,115 @@
...
+    UnknownInstruction,
+    ValueStackHeightMismatch,
+    ValueStackUnderflow,
+    ValueStackUnexpected,
+};
+
+std::string_view to_string(ValidationError);
+
+// https://webassembly.github.io/spec/core/valid/instructions.html#instruction-sequences
+tl::expected<bool, ValidationError> is_valid(CodeEntry const &ce, Module const &m);
```
robinlinden, line 111: No need for argument names in declarations when they don't add anything.

robinlinden (changes requested):
Looks good. A few minor nitpicks, but I think the validation itself looks good!

wasm/validation.cpp
Zer0-One: The alias makes sense, but the helper function seems kinda pointless.

util/variant.h
```diff
@@ -11,6 +11,11 @@
 
 namespace util {
 
+template<typename... T, typename... Ts>
+constexpr bool holds_any_of(std::variant<Ts...> const &v) noexcept {
+    return (std::holds_alternative<T>(v) || ...);
```
robinlinden, line 16: Can you add some unit tests to variant_test.cpp so this is tested even if wasm/... changes?

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+    if (value_stack.size() == control_stack.back().stack_height && control_stack.back().unreachable) {
+        return Unknown{};
+    }
+
+    if (value_stack.size() == control_stack.back().stack_height) {
+        return tl::unexpected{ValidationError::ValueStackUnderflow};
+    }
+
+    auto val = value_stack.back();
```
robinlinden, line 170: Can value_stack be empty here?

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+        if (l.max > k || l.max < l.min) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#block-types
+constexpr bool is_valid(wasm::instructions::BlockType &bt, Module const &m) {
```
robinlinden, line 75: `BlockType const &`

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+    if (!m.type_section.has_value()) {
+        return false;
+    }
+
+    if (auto *t = std::get_if<TypeIdx>(&bt.value)) {
```
robinlinden, line 80: `auto const *`

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+}
+
+tl::expected<ValueOrUnknown, ValidationError> Validator::pop_val_expect(ValueOrUnknown const &expected) {
+    auto actual = pop_val();
+
+    if (!actual.has_value()) {
+        return tl::unexpected{actual.error()};
+    }
+
+    if (actual != expected && !std::holds_alternative<Unknown>(*actual) && !std::holds_alternative<Unknown>(expected)) {
```
robinlinden, line 184: Maybe a separate `if (std::holds_alternative<Unknown>(*actual) || std::holds_alternative<Unknown>(expected)) { return actual; }` before this would be clearer?

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+        }
+
+        popped.insert(popped.begin(), *maybe_val);
+    }
+
+    return popped;
+}
+
+void Validator::push_ctrl(
+        Instruction const &i, std::vector<ValueType> const &params, std::vector<ValueType> const &results) {
```
robinlinden, line 218: These are more suitable to pass by value and move into the `ControlFrame` since they're either heap-allocated or can contain heap-allocated objects.

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+        push_vals(params);
+    }
+}
+
+tl::expected<ControlFrame, ValidationError> Validator::pop_ctrl() {
+    if (control_stack.empty()) {
+        return tl::unexpected{ValidationError::ControlStackEmpty};
+    }
+
+    auto frame = control_stack.back();
```
robinlinden, line 231: Does this need to copy the frame? It contains heap-allocated objects, so a `const &` would be nicer if possible.

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+    if (value_stack.size() != frame.stack_height) {
+        return tl::unexpected{ValidationError::ValueStackHeightMismatch};
+    }
+
+    control_stack.pop_back();
+
+    return frame;
+}
+
+std::vector<ValueType> const &Validator::label_types(ControlFrame &frame) {
```
robinlinden, line 250: `ControlFrame const &`

wasm/validation.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+void Validator::unreachable() {
+    value_stack.resize(control_stack.back().stack_height);
+
+    control_stack.back().unreachable = true;
+}
+
+// TODO(dzero): Serialize operand stack and control stack as part of the ValidationError to make debugging easier
+// https://webassembly.github.io/spec/core/valid/instructions.html#instruction-sequences
+tl::expected<bool, ValidationError> is_valid(CodeEntry const &ce, Module const &m) {
```
robinlinden, line 266: Looks like this can only ever return `true` as the bool? Maybe `expected<void, ValidationError>` is a more suitable return type? I think that's pretty common for "returns nothing or fails"-functions.

robinlinden (changes requested):
First pass, but pretty large PR so will look through the instruction validation more properly soon.

wasm/validation.cpp
Zer0-One: I think that's only possible if a control frame somehow records the wrong stack height when it's entered. I can add an assert to be sure, because I'm not 100% certain.

wasm/validation.cpp
Zer0-One: I think it reads better the way it's written now. "If the popped value wasn't what you expected, and neither the actual or expected values are Unknown, error".

wasm/validation.cpp
Zer0-One: Didn't realize this was even possible.

wasm/validation.cpp
```diff
@@ -0,0 +1,525 @@
...
+                    return tl::unexpected{ValidationError::TypeUndefined};
+                }
+
+                params = m.type_section.value().types.at(*idx).parameters;
+                results = m.type_section.value().types.at(*idx).results;
+            }
+
+            // The case of an empty block type is handled implicitly by leaving the vectors empty
+
+            v.push_ctrl(Block{}, params, results);
```
robinlinden, line 443: `params` and `results` can be moved here and in the `Loop` case.

wasm/validation.cpp
```diff
@@ -0,0 +1,525 @@
...
+            tl::expected maybe_vals =
+                    v.pop_vals(v.label_types(v.control_stack.at(v.control_stack.size() - (branch_if->label_idx + 1))));
+
+            if (!maybe_vals.has_value()) {
+                return tl::unexpected{maybe_vals.error()};
+            }
+
+            v.push_vals(v.label_types(v.control_stack.at(branch_if->label_idx)));
+        } else if (std::holds_alternative<Return>(inst)) {
+            tl::expected maybe_vals = v.pop_vals(v.label_types(v.control_stack.at(0)));
```
robinlinden, line 509: Is control_stack guaranteed to have at least one value on it here? If so, maybe add an assert to make that clearer?

robinlinden (changes requested):
Missing tests for parts of the validation, but otherwise it looks good!

wasm/validation.cpp
Zer0-One: the control stack is guaranteed to always have 1 frame on it, until you're done validating all instructions. but i make that assumption in other places aside from just this one, and there might be more of them when we add the rest of the instructions, so i'll add an assert at the beginning of the loop.

wasm/validation_test.cpp
```diff
@@ -0,0 +1,232 @@
...
+        a.expect(is_valid(m).has_value());
+    });
+
+    s.add_test("Function: invalid trivial sequence", [&](etest::IActions &a) {
+        ce.code = {I32Const{42}, I32Add{}};
+
+        tl::expected res = is_valid(m);
+
+        a.expect(!res.has_value());
+        a.expect_eq(res.error(), ValidationError::ValueStackUnderflow);
```
robinlinden, line 45: Not necessary, but you can write all of this as just `a.expect_eq(is_valid(m), tl::unexpected{ValidationError::ValueStackUnderflow});`. No need to store, check has_value, and unpack the error every time. Comparisons with tl::expected are implemented in a pretty sane fashion.

robinlinden (changes requested):

wasm/validation.cpp
```diff
@@ -0,0 +1,599 @@
...
+        if (!maybe_vals.has_value()) {
+            return tl::unexpected{maybe_vals.error()};
+        }
+    }
+
+    return {};
+}
+
+tl::expected<void, ValidationError> is_valid_funcs(Module const &m) {
+    assert(m.function_section.has_value());
```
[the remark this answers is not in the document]
Zer0-One: It doesn't assert what the validation requires. It asserts what I already checked in the wrapper functions. They're only there to shut up the static analyzer.

Also, I was passing individual sections previously, and then realized that I would end up passing the function section, type section, code section, memory section, and probably more as we add more instructions. Not a pretty solution.

wasm/validation.cpp
```diff
@@ -0,0 +1,599 @@
...
+
+    return true;
+}
+
+// https://webassembly.github.io/spec/core/valid/types.html#block-types
+constexpr bool is_valid(wasm::instructions::BlockType const &bt, Module const &m) {
+    assert(m.type_section.has_value());
+
+    if (auto const *t = std::get_if<TypeIdx>(&bt.value)) {
+        //        if (!m.type_section.has_value()) {
```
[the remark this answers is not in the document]
Zer0-One: Accidentally leftover from testing changes.

wasm/validation.cpp
```diff
@@ -0,0 +1,599 @@
...
+            }
+
+            v.push_vals(maybe_frame->results);
+        } else if (Branch const *branch = std::get_if<Branch>(&inst)) {
+            if (v.control_stack.size() < branch->label_idx) {
+                return tl::unexpected{ValidationError::LabelInvalid};
+            }
+
+            tl::expected maybe_vals =
+                    v.pop_vals(v.label_types(v.control_stack[v.control_stack.size() - (branch->label_idx + 1)]));
```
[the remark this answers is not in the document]
Zer0-One: The check above should have been `<=` instead of `<`

wasm/validation.cpp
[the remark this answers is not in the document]
robinlinden: I get that the asserts aren't required by the validation, but things like the module having a function and a type section are still part of the API of the function, and it's nicer if that's obvious at the call-site.

How about something like
```cpp
FunctionValidator v{*m.function_section, *m.type_section, ...};
for (auto func_idx : idxs) { v.validate(func_idx); }
```
or similar? That way you're still forced to call the function in a way that won't require any asserts and still don't have to pass sections as arguments down the call stack.

wasm/validation.cpp
```diff
@@ -0,0 +1,594 @@
...
+namespace {
+// TODO(dzero): Serialize operand stack and control stack as part of the ValidationError to make debugging easier
+// https://webassembly.github.io/spec/core/valid/instructions.html#instruction-sequences
+tl::expected<void, ValidationError> validate_function(std::uint32_t func_idx, Module const &m) {
+    assert(m.function_section.has_value());
+    assert(m.type_section.has_value());
+    assert(m.code_section.has_value());
+    assert(func_idx < m.function_section->type_indices.size());
+    assert(func_idx < m.code_section->entries.size());
+    assert(m.function_section->type_indices[func_idx] < m.type_section->types.size());
```
robinlinden, line 283: Can the validation of these bounds be moved in here instead of being done one level up in the call stack? That way we won't need these asserts.

robinlinden (commented):
Looks like this is pretty much in shape to be merged. Last nitpicky comments this time, promise. :P

Zer0-One:
it's fixed, review again

robinlinden (approved):

PR #903, opened 2024-03-21
gfx/sfml: Don't try to use fonts that don't support ASCII

robinlinden:


robinlinden:
This is a temporary solution and will be replaced once our font handling is a bit smarter, but getting regular text working working alright is higher priority than emoji-fonts right now.

Zer0-One (approved):

PR #912, opened 2024-03-30
url: Merge BaseParser into UrlParser

robinlinden:
BaseParser was only used in //url and most of its methods were only
tested via UrlParser.


Zer0-One:
You don't think we might use something like BaseParser again in the future? It's gonna be annoying to separate it out again if it does come up.

robinlinden:
Probably not. Sharing it between the different parsers was causing more work than it gave benefit, so all other parsers have also already copied the BaseParser-bits they were using. The functionality was trivial and most of it only used in the parser being worked on when it was added.Not having to think about what effects touching the BaseParser might have on HTML/CSS/URL parsing when working on JS parsing is nice enough that I never want to see BaseParser again.On Mar 30, 2024 03:03, David Zero ***@***.***> wrote:
You don't think we might use something like BaseParser again in the future? It's gonna be annoying to separate it out again if it does come up.

Zer0-One (approved):

PR #919, opened 2024-04-06
protocol/http: Include query-parameters in requests

robinlinden:
Happy now, @GrayHatter?


GrayHatter:
yes, very

:blush: :heart:

PR #920, opened 2024-04-07
Improve rendering of the nginx default error pages

robinlinden:
We were missing the `display: block` for `<center>`, and we had no CSS at all for `<hr>`. We were also dropping any elements that had either height or width == 0, even if a border size would cause things to be rendered anyway.

We're still not centering things inside of the `<center>` element, but that's a future someone's problem.

Before:
image

After:
image

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -82,28 +82,15 @@ wbr {
   display: inline;
 }
 
-address,
-blockquote,
 details,
-dialog,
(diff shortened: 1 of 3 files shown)


GrayHatter (approved):
LGTM

proper support for various "server crashed" error pages is required for compatibility with srctree!

PR #938, opened 2024-04-27
idna: Process ucd code point decomposition data into a C++ish table

robinlinden:
This is needed for the NFC normalization we have to perform when doing the UTS46 processing.


robinlinden:
With this done, I think NFC normalization (built on top of the NFD normalization) is the only real thing left (aside from putting it all together) before we can replace icu with nihcu.

Zer0-One (approved):

PR #949, opened 2024-05-07
type/sfml: Add a workaround for non-breaking space measurements being 0

robinlinden:
This issue can be seen on https://seirdy.one/

Before:
image

Note that the text box containing the nbsp-character has 0 width:
image

After:
image

### DIFF
--- a/type/sfml.cpp
+++ b/type/sfml.cpp
@@ -15,6 +15,7 @@
 #include <SFML/System/String.hpp>
 
 #include <algorithm>
+#include <cstdint>
 #include <filesystem>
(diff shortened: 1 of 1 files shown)


Seirdy:
Oh wow, it's me! Thanks for the fix.

robinlinden:
It is you! You mentioned Hastur on your webpage, so I figured I'd check if there was any low-hanging fruit I could fix to make it work significantly better. :P

Both the `&nbsp;` and `prefers-color-scheme` media query bits were done while looking at your website!

PR #957, opened 2024-05-14
ci: Drop support for Clang 14, 15

robinlinden:
The Ubuntu 24.04 LTS has been released, and GitHub has started rolling out workers for it, so good time to drop some of the older compilers we need workarounds for or feel restricted by.

We'll probably drop support for more things than Clang 15 and older later, but it's a decent start that unblocks a lot of std::ranges-bits and emplace-back optimizations we've had to avoid.


robinlinden:
I haven't shifted things over to Ubuntu 24.04 runners yet as they're still in beta, but I'll update my gcc-14 branch to them and we can run that for a bit to check that the availability is fine.

> The image is marked as "beta" for now. It means some software can be unstable on the new platform. Also, there could be queueing issues as the capacity will be balanced only throughout the next weeks.

Zer0-One (approved):

PR #963, opened 2024-05-19
Calculate prefers-color-scheme media queries based on the OS dark mode preference

robinlinden:
The default is light mode, but you can force it to dark mode by setting the `HST_DARK_MODE` env var to `1`.

Working out if the user wants dark mode or not on Linux is a bunch of work, so that's a todo for someone in the future.

### DIFF
--- a/engine/engine.h
+++ b/engine/engine.h
@@ -27,6 +27,7 @@ namespace engine {
 struct Options {
     // Default chosen by rolling 1d600.
     int layout_width{600};
+    bool dark_mode{false};
 };
 
 struct PageState {
--- a/engine/engine.h
+++ b/engine/engine.h
@@ -24,6 +24,11 @@
 
 namespace engine {
 
+struct Options {
+    // Default chosen by rolling 1d600.
+    int layout_width{600};
+};
+
 struct PageState {
     uri::Uri uri{};
     protocol::Response response{};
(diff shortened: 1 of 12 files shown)


engine/engine.h
```diff
@@ -24,6 +24,12 @@
 
 namespace engine {
 
+struct Options {
+    // Default chosen by rolling 1d600.
```
GrayHatter, line 28: just making sure, you've rolled it a half dozen times then too the mean right?

engine/engine.h
```diff
@@ -24,6 +24,12 @@
...
+    int layout_width{600};
+    bool dark_mode{false};
```
GrayHatter, line 30: I heard light will attract bugs... maybe dark mode should be the default, and users can force enable bright mode?

GrayHatter: Or: this breaks my work flow, I've configured my audio player to read the average power value of each pixel on my monitor, and it detects the power spike as not using the computer and pauses my music. Just add an option to enable light mode. 

There are probably children out there using dark mode to hide from monsters! YOUR UPDATE MURDERS CHILDREN!

GrayHatter (approved):

engine/engine.h
robinlinden: So I considered the problem with the children and decided that they're holding the browser wrong and only have themselves to blame.

That's just the default if a value isn't passed on by the browser, e.g. in the tui where it doesn't matter right now.

The actual default when running the browser is trying to autodetect what the user's theme is and going with that, and then falling back to light mode if unable to detect it. I want dark mode by default, but it's not autodetected on Linux at all, so the default will be light so I'm motivated to improve this.

For now, `export HST_DARK_MODE=1` if you want to use the sane version.

PR #970, opened 2024-05-24
uri: Fix crashes with large uris in combination w/ base uris

robinlinden:


robinlinden:
I'll split out the fuzz test improvements to a separate PR as they're nice but not required for the bug fix. The fuzz test changes introduce some questions about test coverage and unit test handling that I need to look into before merging them.

PR #976, opened 2024-06-02
archive: Add brotli decoding

Zer0-One:


archive/brotli.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <cstddef>
+#include <cstdint>
+#include <span>
+#include <string_view>
+#include <vector>
+
+namespace archive {
+
+enum class BrotliError : std::uint8_t {
```
GrayHatter, line 18: This name is kinda stepping on Brotli's namespace isn't it? Can we use ArchiveError here so it could be reused?

That's an actual question, I'm not sure what the desired level of abstraction around 3rd party libs should be?

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
...
+    std::unique_ptr<BrotliDecoderState, decltype(&BrotliDecoderDestroyInstance)> br_state(
+            BrotliDecoderCreateInstance(nullptr, nullptr, nullptr), BrotliDecoderDestroyInstance);
+
+    if (br_state == nullptr) {
+        return tl::unexpected{BrotliError::DecoderState};
+    }
+
+    // Cap output buffer at 1GB. If we hit this, something fishy is probably
+    // going on, and we should bail before we OOM.
+    std::size_t constexpr kMaxOutSize = 1000000000;
```
GrayHatter, line 50: This isn't 1GB. But it's also seems totally reasonable to have the output >= 1GB.

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (out.size() >= kMaxOutSize) {
+            return tl::unexpected{BrotliError::MaximumOutputLengthExceeded};
+        }
+
+        res = BrotliDecoderDecompressStream(br_state.get(), &avail_in, &next_in, &avail_out, &next_out, &total_out);
+
+        // Because we provide the whole input up-front, there's no reason we
+        // would ever block on needing more input, except for corrupt data
+        if (res == BROTLI_DECODER_RESULT_NEEDS_MORE_INPUT) {
+            return tl::unexpected{BrotliError::InputCorrupt};
```
GrayHatter, line 76: InputIncomplete is the correct error here

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        if (res == BROTLI_DECODER_RESULT_ERROR) {
+            // Brotli doesn't expose this in a sane way, so we use magic
```
GrayHatter, line 80: I had to look it up, first glance, looks like they generate the list using MACRO... first ew... second, this is only correct for as long as they deem it so, which means this might need to be version locked depending on how important this range is.

archive/brotli.h
robinlinden: Brotli's in C and this lives in a C++-namespace, so it should be fine. We'll probably want a unified interface towards the `//archive` stuff at some point, but for now each library has its own set of errors with nothing holding them together other than all of them having the same `expected<vector<byte>, LibError> lib_decode(span<byte const>)`-API.

archive/brotli.cpp
Zer0-One: 1. That is definitely 1GB. If it's not, then I need you to explain how.
2. I can't imagine a scenario where you legitimately need to decode a 1GB of content from a single HTTP response, so I need you to explain this one too.

archive/brotli.cpp
Zer0-One: It's ugly, but at the very least, the only damage that can likely occur from version changes is that we report the wrong error. It'll still return some error when it fails, just might not be the right one.

archive/brotli.cpp
Zer0-One: Incomplete input is corrupt input.

archive/brotli.h
Zer0-One: Once this is merged, we'll be good on content encoding; I'm not sure we need to bother with lzw at all. But I think this is a concern for *after* this PR is merged, not before. Can add an issue to the tracker for it and assign it to me if you want.

robinlinden: lzw is super rare when it comes to content encodings, but I'm messing around with it in other contexts, so it'll be ~free for us to support if we need it elsewhere anyway.

I think it's better to hold off on a unified archive API for now. I'm messing around with making the engine bits a bit more modular so that users of it can add their own or opt out of the default stuff that we have right now, e.g. for experimenting with adding more image encodings or transfer encodings via some form of plugin-API. Might change how we use the archive bits, so tying all archive-implementations together before we see what happens there doesn't feel worthwhile.

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
+// SPDX-FileCopyrightText: 2024 David Zero <zero-one@zer0-one.net>
+//
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include "archive/brotli.h"
+
+#include "brotli/decode.h"
```
robinlinden, line 7: `<>` for external libraries.

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
...
+    std::vector<std::byte> out;
+
+    std::size_t avail_in = input.size();
+    auto const *next_in = reinterpret_cast<std::uint8_t const *>(input.data());
+    std::size_t total_out = 0;
+
+    BrotliDecoderResult res = BROTLI_DECODER_RESULT_ERROR;
+
+    while (res != BROTLI_DECODER_RESULT_SUCCESS) {
+        std::vector<std::byte> intermediate_buf(kChunkSize);
```
robinlinden, line 62: Creating this here means a new allocation for every chunk. If you create it outside the while-loop scope, the same allocation can probably be reused for all chunks.

archive/brotli.cpp
```diff
@@ -0,0 +1,101 @@
...
+            return tl::unexpected{BrotliError::BrotliInternalError};
+        }
+
+        intermediate_buf.resize(kChunkSize - avail_out);
+
+        // TODO(zero-one): Replace with insert_range() when support is better
+        out.insert(out.end(), intermediate_buf.begin(), intermediate_buf.end());
+    }
+
+    out.resize(total_out);
```
robinlinden, line 96: Is this actually needed? `out` starts w/ a size of 0 and after decompressing a chunk, we insert only the newly decompressed bits `kChunkSize - avail_out`?

robinlinden (commented):
Looks good, just some C++-y nitpicking as usual. Great job on the test coverage. Looks like you hit pretty much every error condition possible with the current API. :P

robinlinden:
The clang-cl error is just git_repository being so slow it times out (which is the reason I've avoided it so far). I've hit it too and it's on my TODO to fix that being flaky. A rerun should fix it.

archive/brotli.cpp
GrayHatter: They're unique states. In one, more data would fix the problem, and allow it to continue, in the other more data won't fix the problem.

archive/brotli.cpp
GrayHatter: Bah! you're correct, I got GiB and GB confused... again :<

GrayHatter: 2) 
this is terminal output from seconds ago
```
grayhatter@host ~ $ rsync -vPz ~/[redacted/server-local.zip gr.ht:/srv/http/gr.ht/[redacted].zip
[redacted]-server-local.zip
    7,322,472,876 100%    5.01MB/s    0:01:01 (xfr#1, to-chk=0/1)

sent 7,320,562,208 bytes  received 225,626 bytes  5,050,202.11 bytes/sec
total size is 7,322,472,876  speedup is 1.01

robinlinden (approved):

PR #978, opened 2024-06-03
style: Match classes attached to a specific element type

robinlinden:
Before:
before

After:
after


robinlinden:
Freedesktop's GitLab is down again. We really need to sort out mirroring for the one freedesktop repo (xcursor) we have without a mirror.

>GitLab will be down for maintenance Monday June 03, from approx 10am UTC for roughly 48 hours.

image

PR #979, opened 2024-06-04
archive: Split into one target per archive format

robinlinden:
This both makes the coverage feedback in fuzzing nicer, and allows things like the GIF image parser to only depend on lzw without also pulling in gzip/zlib-handling.

### DIFF
--- a/archive/BUILD
+++ b/archive/BUILD
@@ -3,14 +3,9 @@ load("@rules_fuzzing//fuzzing:cc_defs.bzl", "cc_fuzz_test")
 load("//bzl:copts.bzl", "HASTUR_COPTS", "HASTUR_FUZZ_PLATFORMS")
 
 cc_library(
-    name = "archive",
-    srcs = glob(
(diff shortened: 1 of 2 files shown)


Zer0-One (approved):
LGTM, but I don't think there's a point in separating zlib and gzip, since it's exactly the same code and same API.

robinlinden:
> LGTM, but I don't think there's a point in separating zlib and gzip, since it's exactly the same code and same API.

It's more that we'll probably want a more unified API in front of our different archive formats in the future when we have more of an idea of what kind of options we want for things like e.g. limiting the output size.

`zlib_decode(data, ZlibMode::Gzip)` is not as nice as something like `Gzip::decode(data)` or `gzip_decode(data)`, and no other decode API will take a ZlibMode argument, so we're sort of leaking internal mechanics instead of hiding them.

Anyway, that stuff is just ideas right now, so nothing I'm touching any time soon. :P

PR #994, opened 2024-06-22
html+html2: Implement enough of the new parser to drop the old one

robinlinden:


robinlinden:
This is on hold until some issues with whitespace-collapsing have been fixed. The old HTML parser would drop all elements that only contained whitespace, which is obviously not correct in cases like `<span>hello</span> <span>world</span>`, but sort of correct in cases like `<p>hello</p>\n\n<p>world</p>`. Something that deals with this more properly will be added to the whitespace-collapsing system to not break too much when this is merged.

PR #998, opened 2024-06-26
archive: Use `std::byte` as the input type in the API

robinlinden:
std::byte is like the char types in that it's allowed to alias anything.

std::uint8_t doesn't have this magic, so when using it we also have to
static_assert that it's an alias of one of the blessed character types
that can alias anything.

The zlib-decoding used char types, so it didn't have this issue, but it
was also updated for consistency.

### DIFF
--- a/archive/gzip_fuzz_test.cpp
+++ b/archive/gzip_fuzz_test.cpp
@@ -1,18 +1,17 @@
-// SPDX-FileCopyrightText: 2023 Robin Lindén <dev@robinlinden.eu>
+// SPDX-FileCopyrightText: 2023-2024 Robin Lindén <dev@robinlinden.eu>
 //
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include "archive/zlib.h"
 
+#include <cstddef>
 #include <stddef.h> // NOLINT
 #include <stdint.h> // NOLINT
-#include <string_view>
 #include <tuple>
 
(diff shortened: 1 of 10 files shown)


Zer0-One (approved):

Zer0-One (commented):
It looks like std::byte is an alias of `unsigned char`, not a fixed-width type. The decode APIs for zstd and brotli both expect 8-bit bytes and use uint8_t; is there any scenario where using std::byte instead of a fixed-width type would std::byte us in the ass? e.g, a platform with 16-bit char? And if so, maybe it would be worth adding a `static_assert` at the namespace level that fails if `CHAR_BIT != 8`?

robinlinden:
Ah, hmm. Good point. I think the static assert is a good idea, so I'll add one in the zstd decoder to ensure that we don't break that by accident. Looks like zlib doesn't need it as its input-type already is an `unsigned char`.

>would std::byte us in the ass

lol

robinlinden:
@Zer0-One Happy with a static_assert like this? I figured maybe it makes sense to place it close to where it's moved from our side where we can support any-number-of-bits-bytes to the side where we changing things is high-effort.
https://github.com/robinlinden/hastur/blob/9b10932ab5490cd2eb55aa4e1f2824d8a4c42d24/archive/zstd.cpp#L55-L60

Zer0-One (approved):

PR #1002, opened 2024-06-27
Reduce the use of libfmt for debug print formatting

robinlinden:
This makes for easier switching to std::format later.

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -187,8 +187,9 @@ std::unique_ptr<type::IType> create_font_system() {
             }
         }
 
-        spdlog::warn(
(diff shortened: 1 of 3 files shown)


robinlinden:
Guess maybe this isn't feasible yet. libc++ doesn't define `__cpp_lib_format` to indicate that it supports formatting, even though cppreference says it's implemented as of libc++17. :(

Will drop the switching to std::format in spdlog and keep the removal of libfmt in places where it's not really needed.

See: `https://github.com/llvm/llvm-project/issues/77773`

robinlinden:
This isn't worth it. Compiler support isn't good enough to drop libfmt yet without too much faff.

PR #1021, opened 2024-08-12
wasm: Fix variable-instruction validation

robinlinden:


Zer0-One (approved):
Good catch. Should have labeled those functions `nodiscard` originally.

robinlinden:
> Good catch. Should have labeled those functions `nodiscard` originally.

Thanks! I missed it in review too, but I noticed when going over and updating things when testing std::expected and actually had to touch these lines.

PR #1022, opened 2024-08-15
deps/wpt: Work around git_repository timing out in CI

robinlinden:
Using an older commit of wpt is unfortunate, but git_repository times out a lot when running the build in CI, and commits after this fail the integrity check by having archives where the content changes every now and then.


robinlinden:
I'll rerun this once things have shifted a bit in wpt just to double check that I'm not breaking the integrity check again, but this commit should be fine.

Zer0-One (approved):

PR #1057, opened 2024-09-28
ci: Decrease timeouts for jobs that should be ~instant

robinlinden:
With the exception of link-liveness which takes ~30s, all of these jobs take max ~10s, so having a 30 minute timeout feels a bit silly.


robinlinden:
codecov/project is red because something changed somewhere again and now partially covered lines are reported again.

PR #1127, opened 2024-12-13
build: Enable bzlmod

robinlinden:
This only adds the dependencies that are required for Bazel to be happy
with not being able to access a bzlmod registry, and no dependencies are
updated in this change.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,296 @@
+""" This comment is required for Buildifier to be happy. """
+
+module(name = "hastur")
+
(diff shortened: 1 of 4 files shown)


MODULE.bazel
Zer0-One, line 1: Don't want to call `module()` at the top here?

robinlinden: Good point. I didn't think about someone else consuming this repo as a module at all. Added.

Zer0-One (approved):

PR #1154, opened 2025-01-14
Set up a nih JSON library

robinlinden:


Zer0-One (approved):

robinlinden:
There's some work remaining (replacement characters instead of failing parsing on invalid data, I think?) before we can switch the url wpt test to this lib, but I'll sort that out and make the error messages nicer after merging this initial version. :P

PR #1168, opened 2025-01-27
archive: Make the maximum output size configurable

robinlinden:


Zer0-One:
Why wrap the decoder in a class instead of just adding a maximum length parameter with a default value so that it can be omitted?

robinlinden:
It can be omitted with the class as well. The reason for the class is that I want to be able to pass around configured decoders to make the engine setup a bit more flexible in the future.

Zer0-One (approved):

PR #1173, opened 2025-02-05
wasm: Add validation for table section

Zer0-One:


robinlinden (approved):
Can you also add a test with a valid table? Right now we only test 0 tables, or invalid tables.

PR #1199, opened 2025-03-06
ci: Update codecov-cli

robinlinden:
The tokenless upload setup was apparently redone in codecov-cli 0.9.0,
so there's a chance a non-ancient version will work in forks now.


robinlinden:
@Zer0-One I *think* this should work in forks now (`https://github.com/codecov/codecov-cli/pull/533`), but if it breaks anything for you, let me know and we'll try the `codecov/codecov-action`-method instead as I saw they had some different form of tokenless upload support at least a while back (`https://github.com/codecov/codecov-action/blob/e1b169f2838f7e502ad9f222fbc6ed3e2594e003/src/buildExec.ts#L50-L54`).

robinlinden:
I switched this repo to using tokenless uploads as well now, so 90% sure it won't break PRs. :P

Zer0-One (approved):

PR #1214, opened 2025-03-21
style: Don't log successful var-fallback lookups

robinlinden:
This was incredibly noisy on websites that make use of `var(--foo, 123px)`.


robinlinden:
Merged as 46aab9a787f0f195dab6cf25e94b25e3ccc87bf4.

PR #1242, opened 2025-04-19
util/string: Switch split_once to take a char delimiter

robinlinden:
The only place we used a multi-char delimiter was in a unit test, so no real need to support that use-case.

### DIFF
--- a/css/parser.cpp
+++ b/css/parser.cpp
@@ -687,7 +687,7 @@ void Parser::expand_border_radius_values(Declarations &declarations, std::string
     std::string top_right;
     std::string bottom_right;
     std::string bottom_left;
-    auto [horizontal, vertical] = util::split_once(value, "/");
+    auto [horizontal, vertical] = util::split_once(value, '/');
     Tokenizer tokenizer(horizontal, ' ');
 
     auto unchecked_get = [](Tokenizer &t) -> std::string_view {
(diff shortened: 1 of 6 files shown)


Zer0-One (approved):
PR is ok, but IMHO we should just leave `split_once()` as-is. If we want a multi-char delimiter later, we'll either have to go back and add an overload or revert all these little changes.

robinlinden:
> PR is ok, but IMHO we should just leave `split_once()` as-is. If we want a multi-char delimiter later, we'll either have to go back and add an overload or revert all these little changes.

It's trivial to add back, and even though we've had support for this for ~3 years, we're not using it. No need to keep the inefficient way of doing things around imo. The original reason for this PR is that the stdlib has a more efficient char-overload of things like `find`, and it's enough of a difference between the overloads that clang-tidy has a warning telling you to switch to the `char` one when possible.

Zer0-One:
> > PR is ok, but IMHO we should just leave `split_once()` as-is. If we want a multi-char delimiter later, we'll either have to go back and add an overload or revert all these little changes.
> 
> It's trivial to add back, and even though we've had support for this for ~3 years, we're not using it. No need to keep the inefficient way of doing things around imo. The original reason for this PR is that the stdlib has a more efficient char-overload of things like `find`, and it's enough of a difference between the overloads that clang-tidy has a warning telling you to switch to the `char` one when possible.

Ah, I see. Makes sense.

PR #1245, opened 2025-04-24
Use an embedded fallback font

robinlinden:
We used to look really hard, and die if we couldn't find a font. Now there's always a font available.

There was no thought behind picking Open Sans over any other font, so we can switch to whatever's nice later.


Zer0-One (approved):

robinlinden:
Renamed the "last_resort" things to "fallback". It's a name that stuck around from me trying the unicode last-resort font early when working on this. :P

PR #1250, opened 2025-04-29
wasm: Add validation for start section

Zer0-One:


wasm/validation.cpp
```diff
@@ -682,6 +686,33 @@ tl::expected<void, ValidationError> validate(Module const &m) {
...
 
+    // https://webassembly.github.io/spec/core/valid/modules.html#start-function
+    if (m.start_section.has_value()) {
+        // Ensure function and type sections exist
+        if (!m.function_section.has_value()) {
+            return tl::unexpected{ValidationError::FunctionSectionUndefined};
+        }
+
+        if (!m.type_section.has_value()) {
+            return tl::unexpected{ValidationError::TypeSectionUndefined};
```
robinlinden, line 697: This branch isn't reached in the test intended to test it. Some earlier check must fail before we get here in the test intended to check this. Maybe we should (in the future, in a separate PR) return more detailed error info so you know what section validation failed at if it fails. :P

I think this is reachable if you have an empty function section when calling validation, so the test should be updated so this branch is hit.

wasm/validation.cpp
```diff
@@ -682,6 +686,33 @@ tl::expected<void, ValidationError> validate(Module const &m) {
...
+        }
+
+        // Ensure start function has valid index into function and type section
+        if (m.start_section->start >= m.function_section->type_indices.size()) {
+            return tl::unexpected{ValidationError::StartFunctionInvalid};
+        }
+
+        if (m.start_section->start >= m.type_section->types.size()) {
+            return tl::unexpected{ValidationError::StartFunctionTypeInvalid};
```
robinlinden, line 706: This isn't hit by the tests. Can you add a test w/ an empty type section? I /think/ that would hit this..

robinlinden (approved):

Zer0-One:
ok, so version 3 of the spec is completely different, and most of the work i did is effectively obsoleted... this is a bother :(

PR #1261, opened 2025-05-12
url: Fix ascii alphanumerics being considered invalid url code points

robinlinden:


url/url.cpp
```diff
@@ -1682,6 +1682,10 @@ std::optional<std::string> UrlParser::parse_opaque_host(std::string_view input)
 }
 
 bool UrlParser::is_url_codepoint(std::uint32_t cp) const {
+    if (unicode::is_ascii(cp) && util::is_alphanumeric(static_cast<char>(cp))) {
```
Zer0-One, line 1685: `is_ascii()` looks redundant here.

Zer0-One (dismissed):

url/url.cpp
robinlinden: It's very important actually, that's why I added it. :P We're working with code points here, and lots of u32s cast to chars will have the same value.

```cpp
int main() {
    std::uint32_t cp1 = 'A';
    std::uint32_t cp2 = cp1 | (0b1111'1111 << 8);
    auto c1 = static_cast<char>(cp1);
    auto c2 = static_cast<char>(cp2);
    std::println("{}={} {}={} ", cp1, c1, cp2, c2);
}
```
=> `65=A 65345=A`

Zer0-One (approved):

robinlinden:
lmao, yeah, GitHub's very intuitive interface strikes again.

PR #1278, opened 2025-06-01
engine: Hook up the JS tokenizer

robinlinden:
This doesn't run JS yet, but I'll open a PR for those parts soon. This is enough to make it easy to see what's needed for the tokenizer to work on websites we're testing against.

I might break things apart a bit later to allow the option of using a not-nih JS engine, but that's not really in scope while we're still getting all the basics set up.

```
[2025-06-01 19:54:58.344] [W] Failed to tokenize JavaScript in <script> tag:

    document.querySelectorAll("a").forEach((link) => {
      link.addEventListener("mouseover", (ev) => {
        if (link.className == "vid") {
          document.querySelector("#img").src = "";
          document.querySelector("#video").src = link.href;
          document.querySelector("#video").style.display = "revert";
        } else {
          document.querySelector("#img").src = link.href;
          document.querySelector("#video").pause();
          document.querySelector("#video").style.display = "none";
        }
      });
    });
```

### DIFF
--- a/browser/gui/app.cpp
+++ b/browser/gui/app.cpp
@@ -848,6 +848,14 @@ void App::run_debug_widget() {
         }
     }
 
+    if (ImGui::Checkbox("Enable JavaScript", &enable_js_)) {
+        if (maybe_page_) {
+            // Scripting affects the html parsing, so we actually
+            // need to reload the page for now.
+            reload();
+        }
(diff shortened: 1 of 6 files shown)


GrayHatter:
:tada: imgzen soon!

PR #1304, opened 2025-07-27
browser/gui: Set up an about-page handler

robinlinden:


Zer0-One (approved):
I've been waitin for this one

PR #1313, opened 2025-08-15
tool: Set up a tool for embedding resources in a nicer-than-xxd way

robinlinden:
This will go away once we target compilers w/ support for `#embed`, but
I feel this is a decent intermediate step as this is very little code,
allows us to treat the resources as compile-time constants instead of
mutable data, and doesn't require any namespace hacks where we include
the resources.

There's some issue w/ compiling the font embed file in MSVC in CI (without any output), but I can't reproduce it locally. Leaving that one as xxd while I investigate.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -4,7 +4,7 @@ ignore=body-is-missing
 # TODO(robinlinden): Better way of documenting and setting this up.
 # Each commit must start with the main area it affects.
 [title-match-regex]
-regex=^(archive|azm|browser|bzl|css|css2|dom|dom2|engine|etest|geom|gfx|html|html2|idna|img|js|json|layout|net|os|protocol|render|style|tui|type|unicode|uri|url|util|wasm|all|build|ci|deps|doc|meta)(/.*|\+.*)?:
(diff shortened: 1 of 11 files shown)


robinlinden:
> There's also https://github.com/graphitemaster/incbin

At a glance it looks like it might not support wasm? There's an unmerged PR looking at adding support for emscripten though.

It generates the same type of variables as xxd did (separate ptr+size), but it's nice that it's actually const w/ incbin. I think having a span<byte>, array<123, byte>, or string_view is even better though since we're going to have to repack all data into one of those when integrating it into our codebase anyway.

incbin also doesn't generate data that can be `constexpr` in C++, so no compile-time trickery for making things more efficient if we want to do things like that.

No good support for MSVC without hooking it up with genrules the way I did with this tool, so no convenience win there.

I'd rather just use the 60-line (right now, I'm sure it'll change) nih tool for the 8 or so months until we start targeting ubuntu 26.04 and drop a bunch of older compilers and this tool when we start using `#embed`.

Zer0-One (approved):

Issue #1471, opened 2026-05-31
wasm: Complete the WASM interpreter basics


Swiftaxe:
This issue aims to complete the WASM interpreter so it can handle a basic WASM module.

Add support for the following instructions to the WASM interpreter:

- [x] End
- [x] Return
- [x] Call
- [x] i64.const
- [x] i64.eq, i64.lt_s, i64.add, i64.sub, i64.mul
- [x] f32.const
- [ ] i32.mul, i32.div_s, i32.div_u, i32.rem_s, i32.rem_u
- [ ] i32.shl, i32.shr_s, i32.shr_u, i32.rotl, i32.rotr
- [ ] i32.clz, i32.ctz, i32.popcnt
- [ ] i32.eqz, i32.eq, i32.ne, i32.lt_u, i32.gt_u, i32.le_u, i32.ge_u
- [ ] i32.wrap_i64, i32.extend8_s, i32.extend16_s, i32.reinterpret_f32
- [ ] i32.trunc_f32_s, i32.trunc_f32_u
- [ ] block, loop, br, br_if
- [ ] select

Part of #1470

PR #1472, opened 2026-05-31
wasm: Support 'End' instruction in interpreter

Swiftaxe:
Add support for 'End' instruction to WASM interpreter according to the WASM spec: https://webassembly.github.io/spec/core/exec/instructions.html#blocks

Part of #1471

### DIFF
--- a/wasm/interpreter.h
+++ b/wasm/interpreter.h
@@ -96,6 +96,11 @@ public:
         return tl::unexpected{Trap::UnhandledInstruction};
     }
 
+    // https://webassembly.github.io/spec/core/exec/instructions.html#blocks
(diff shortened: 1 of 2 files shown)


robinlinden (approved):
I think this is fine for now, but we may have to revisit how we deal with `End` once we have more proper block support in the interpreter. :P E.g. right now we don't deal with clearing the stack and have no labels to identify nested scopes.

Feel free to add an author line at the bottom of the `SPDX-FileCopyrightText`-list in the copyright header when you make changes you feel are reasonable to do so for.

We've only used rebases to deal with branching so far, so when updating your branches, prefer that. I'll rebase this one, so no need to do anything to remove the merge-commit you have in this branch. :P

Good PR, thank you!

PR #1473, opened 2026-06-01
wasm: Add support for 'Return' in interpreter

Swiftaxe:
Add support for 'Return' in interpreter.
Spec: https://webassembly.github.io/spec/core/exec/instructions.html#returning-from-a-function

As with 'End', implementation of the full spec for 'Return' is relevant first when nested blocks are implemented ('block', 'if', 'loop'). 

Part of #1471

### DIFF
--- a/wasm/interpreter_test.cpp
+++ b/wasm/interpreter_test.cpp
@@ -58,6 +58,20 @@ int main() {
         a.expect_eq(res, std::nullopt);
     });
 
+    s.add_test("A Return instruction exits the function and returns the top-of-stack value.", [](etest::IActions &a) {
+        // https://webassembly.github.io/spec/core/exec/instructions.html#returning-from-a-function
+        Interpreter i;
+        auto res = i.run({{I32Const{42}, Return{}, I32Const{99}}});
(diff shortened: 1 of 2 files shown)


wasm/interpreter_test.cpp
```diff
@@ -58,6 +58,20 @@ int main() {
         a.expect_eq(res, std::nullopt);
     });
 
+    s.add_test("A Return instruction exits the function and returns the top-of-stack value.", [](etest::IActions &a) {
```
robinlinden, line 61: Prefer shorter test titles in the future. :P We get the line anything goes wrong on in the output as well, and long titles both don't get along with terminals and make filtering out what tests to run a bit annoying.

robinlinden (approved):

robinlinden:
Ah, also thank you! Looks good. :D

wasm/interpreter_test.cpp
Swiftaxe: That's a new test naming convention to me, but sure, let's do that. 🤓

PR #1474, opened 2026-06-01
js: Add basic object support

robinlinden:
Also seems fine in the repl, even if parts of the interpreter need tweaking for full support:
```
INFO: Running command line: bazel-bin/js/repl/repl
'/quit' to quit.
> a = { b: 3 };
[Object]
> a.b;
3.000000
```

### DIFF
--- a/js/parser_test.cpp
+++ b/js/parser_test.cpp
@@ -449,5 +449,72 @@ int main() {
         a.expect(std::holds_alternative<js::ast::EmptyStatement>(second_statement));
     });
 
+    s.add_test("object expression, empty", [](etest::IActions &a) {
+        auto p = js::Parser::parse("a = {};").value();
+
+        a.expect_eq(p.body.size(), std::size_t{1});
+        auto &statement = p.body.at(0);
+        auto &expr_stmt = std::get<js::ast::ExpressionStatement>(statement);
+        auto &assign = std::get<js::ast::AssignmentExpression>(expr_stmt.expression);
+        a.expect_eq(std::get<js::ast::Identifier>(*assign.left).name, "a");
+
+        auto &obj_expr = std::get<js::ast::ObjectExpression>(*assign.right);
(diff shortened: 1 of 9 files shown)


js/parser_test.cpp
```diff
@@ -449,5 +449,43 @@ int main() {
         a.expect(std::holds_alternative<js::ast::EmptyStatement>(second_statement));
     });
 
+    s.add_test("object expression, empty", [](etest::IActions &a) {
+        auto p = js::Parser::parse("{};").value();
```
Swiftaxe: Maybe not part of this scope, but asserting '{};' as an object expression is not accurate to spec. '{}' at the statement level is an empty block-level statement. Consider whether something like this `if (true) { x = 1; }` might be parsed as an object literal.

Swiftaxe (approved):
LGTM! 😃

Also tested:

- Multiple values works: `a = { b: 1, c: 2 }`
- Nested objects works: `a = { b: { c: 1 } }`
- String literal fails: `a = { "b": 3 }`
- Numeric literal fails: `a = { 1: 2 }`

The failing ones are perhaps a bit out of scope and could be addressed in a different PR.

js/parser_test.cpp
robinlinden: Whoops, good catch. :D

I was thinking of expressions when writing this, but as you as, this will obviously break once we support blocks when parsing statements. Fixed the tests so the object-expressions now always are treated as expressions and added a test for nested objects.

I didn't even know "labeled statements" were a thing, but apparently `{a: 3}` is parsed as a block statement containing a labeled statement containing a statement-expression when it appears somewhere a statement is fine. What nonsense. :P

robinlinden:
@Swiftaxe Thanks for having a look! :D Will leave non-identifier keys as a future improvement since this already isn't a tiny PR and still moves things in the right direction.

Swiftaxe (approved):
LGTM!

PR #1476, opened 2026-06-07
deps: Add glfw

robinlinden:
This has less extra nonsense than SFML, both in functionality and dependencies. glfw is also usable when compiling to wasm, so if we set something up w/ this, we can in theory run it in a browser as well: https://emscripten.org/docs/compiling/Contrib-Ports.html#contrib-glfw3

I verified that I can run the example on both Windows and Linux, and I also checked that everything builds on macOS, but didn't run anything.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -51,8 +51,6 @@ build:macos      --cxxopt='-std=c++2b'
 build:macos --host_cxxopt='-std=c++2b'
 build:macos      --cxxopt='-fno-rtti'
 build:macos --host_cxxopt='-fno-rtti'
-build:macos --objccopt='-ObjC++'
-build:macos --objccopt='-std=c++2b'
 
 build:windows --enable_runfiles
 build:windows --action_env=LOCALAPPDATA # Quirk for running vswhere, remove when icu no-longer needed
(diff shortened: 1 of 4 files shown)


robinlinden:
rules_python seems to be flaky on Windows. We should look into replacing glad w/ something that doesn't require Python since that's the only reason we need it.
```
FAIL: //third_party:requirements.test (Exit 2) (see C:/users/runneradmin/_bazel_runneradmin/425h7iu3/execroot/_main/bazel-out/x64_windows-fastbuild/testlogs/third_party/requirements.test/test.log)
INFO: From Testing //third_party:requirements.test:
==================== Test output for //third_party:requirements.test:
C:/users/runneradmin/_bazel_runneradmin/425h7iu3/external/rules_python++python+python_3_12_3_x86_64-pc-windows-msvc/python.exe: can't open file 'C:\\Users\\runneradmin\\_bazel_runneradmin\\425h7iu3\\execroot\\_main\\bazel-out\\x64_windows-fastbuild\\bin\\third_party\\requirements.test.exe.runfiles\\_main\\third_party\\requirements.test.zip': [Errno 13] Permission denied
================================================================================
```

Will rerun the jobs and look into that separately.

Zer0-One (approved):

PR #1477, opened 2026-06-10
wasm: Add support for 'Call' in interpreter

Swiftaxe:
## Description:

Adds function call support to the Wasm interpreter.

Introduces a Function struct on Interpreter holding a instruction body and a local count.
interpret(Call) executes the callee in an isolated frame — caller locals and returning state are saved and restored around the call.
The callee's result (if any) is pushed onto the caller's stack, allowing the caller to continue executing with it.

**Three tests cover:* 

- Basic invocation
- Frame isolation (callee writes don't affect caller locals)
- That the caller resumes correctly after the call returns.

**Note:** Suggestions for writing shorter test descriptions that remain understandable are welcome. 😄

### DIFF
--- a/wasm/interpreter.h
+++ b/wasm/interpreter.h
@@ -18,6 +18,7 @@
 #include <optional>
 #include <span>
 #include <tuple>
+#include <utility>
 #include <variant>
 #include <vector>
 
@@ -107,10 +108,18 @@ enum class Trap : std::uint8_t {
 class Interpreter {
 public:
     using Value = std::variant<std::int32_t, std::int64_t, float>;
+
+    struct Function {
+        std::vector<instructions::Instruction> body;
+        std::uint32_t local_count{};
+        [[nodiscard]] bool operator==(Function const &) const = default;
+    };
+
     std::vector<Value> stack;
     std::vector<Value> locals;
     std::vector<Value> globals;
     std::vector<std::uint8_t> memory;
+    std::vector<Function> functions;
     bool returning{false};
     [[nodiscard]] constexpr bool operator==(Interpreter const &) const = default;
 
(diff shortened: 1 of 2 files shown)


Zer0-One:
Could shorten up the test names by doing, e.g: 
```
call: function invocation
```
instead of:
```
A Call instruction invokes a function and pushes its result.
```

As long as the intention is relatively clear, you can shorten most of those up to just 2-3 words each.

wasm/interpreter.h
```diff
@@ -110,6 +118,37 @@ class Interpreter {
         return {};
     }
 
+    // https://webassembly.github.io/spec/core/exec/instructions.html#function-calls
+    // Invoke function[idx] in its own isolated frame; push its result (if any) onto our stack.
+    tl::expected<void, Trap> interpret(instructions::Call const &call) {
+        if (call.function_idx >= functions.size()) {
```
robinlinden: Validation is responsible for making sure that this can't happen, so replace it with a regular C-style `assert(call.function_idx < functions.size());`

wasm/interpreter.h
```diff
@@ -110,6 +118,37 @@ class Interpreter {
...
+            return tl::unexpected{Trap::UnhandledInstruction};
+        }
+        auto const &fn = functions[call.function_idx];
+
+        // Save caller frame.
+        auto saved_locals = locals;
```
robinlinden: Use `auto saved_locals = std::exchange(locals, {})` here. It's both less code and will move the `locals`-member so we don't have to first copy it, then deallocate the one we copied from.

wasm/interpreter_test.cpp
```diff
@@ -72,6 +72,33 @@ int main() {
         a.expect_eq(res, std::nullopt);
     });
```
robinlinden, line 74: Please also add a test for when the called function traps. :P

robinlinden (changes requested):
Nice, great work. :D

wasm/interpreter.h
```diff
@@ -110,6 +118,30 @@ class Interpreter {
...
+        // Restore caller frame.
+        locals = std::move(saved_locals);
+        returning = saved_returning;
+
+        if (!result) {
+            return tl::unexpected{result.error()};
+        }
+        if (result->has_value()) {
+            stack.push_back(result->value());
+        }
```
robinlinden: clang-tidy saying this is unchecked is a bit silly. I think it's related to it being hidden inside a std::expected when we check (because it's an `expected<optional<Value>, Error>` there are two layers that need to be okay for it to be safe to unpack).. You can probably work around it by unpacking one layer at a time:
```
auto &retval = result.value();
if (retval.has_value()) {
  stack.push_back(*retval);
}
```

robinlinden: Ah, and when addressing review comments, please --amend the previous commit instead of creating new ones. :P We want every commit to pass CI and stuff.

Swiftaxe: Sure thing! This preference is different between projects. 🐻‍❄️

robinlinden: Yeah, no worries. :D Sorry for the bother, all of this is undocumented because we've been ~2-3 people working on this forever, but it might be worth writing down the contribution guidelines to make onboarding easier. :P 

Usually I comment on it and sort it out myself when merging PRs the first few times if needed so new people don't feel that undocumented nonsense is too annoying, but now I figured since you're probably fixing the clang-tidy problem anyway..

Swiftaxe: No problem. It's a bit funny, though, as I was recently instructed to do the opposite in another project. Developer opinions differ. 😆

As there is no flood of new contributors at the moment, this process works well. 😁

robinlinden (approved):

PR #1480, opened 2026-06-11
wasm: Add interpreter support for 'i64.const'

Swiftaxe:
Adds i64.const support to the Wasm interpreter.


robinlinden (approved):
Nice!

robinlinden:
I increased the timeouts in the macOS jobs since they keep timing out and rebased this PR, but now they finished way under the original 30min-limit. Oh well, guess GitHub Actions' runners are just a bit flaky/slow today.

PR #1486, opened 2026-06-17
wasm: Add interpreter support for i64 arithmetic

Swiftaxe:
Adds five i64 instructions to the interpreter: i64.eq, i64.lt_s, i64.add, i64.sub, and i64.mul.

This time with short test names. 😉


robinlinden (approved):
Nice! :D

PR #1491, opened 2026-06-20
idna: Support encoding text as punycode

robinlinden:
The punycode stuff still needs better handling for bad input (especially the CodePointView), but this has all the functionality in place. I ran the uts46 spectests and the url wpt tests against it, and they seemed happy with all ascii-punycode-related bits.

I didn't end up needing the constexpr expect and require things in this PR, but it's a tiny change and makes sense on its own.


Zer0-One (approved):
LGTM-with-SUPER-MONKEYS

PR #1498, opened 2026-08-06
all: Migrate from tl::expected to std::expected

robinlinden:
There's some work to do in //browser, //engine, and //js where it wasn't just a matter of replacing the types and includes.


Zer0-One (approved):
I'll be honest, I didn't read the whole diff, but I showed it to my dog and he said it's good.

PR #1520, opened 2026-09-14
build: Simplify our clang-cl setup

robinlinden:


robinlinden:
I guess the CI doesn't really make it obvious that the clang-cl job still builds w/ clang-cl, but I tested this in a VM and it appears to do what I want. Maybe we should think about stamping version and compiler information into the binary in the future?

PR #1524, opened 2026-09-19
Improve CI stability

robinlinden:


Zer0-One:
I had no clue why this one was happening, so I threw it at Fable, and it says:

`Short answer: the -1 is a data race in GCC's coverage counters, not a BuildBuddy or lcov problem. Bazel's coverage build compiles with gcc's default -fprofile-update=single, which is not thread-safe. net/socket_test runs an asio io_context on a second thread while the main thread drives the client, so both threads execute asio's mutex code (conditionally_enabled_mutex.hpp, futex_slim_mutex.hpp) concurrently and race on the same counters. The fix is to compile coverage builds with -fprofile-update=atomic.`

Try the coverage builds with `-fprofile-update=atomic` and see what happens?

robinlinden:
> [...]
> Try the coverage builds with `-fprofile-update=atomic` and see what happens?

I didn't know this was a thing, but read up on it. The default's apparently `-fprofile-update=prefer-atomic` if `-pthread` is on the command line, and it's obviously supposed to be since we're doing multithreading, but it wasn't. :( We were linking pthread via boringssl, and they incorrectly(?) don't pass `-pthread` when compiling, and just link against it.

#1525 should hopefully make the coverage-collection atomic.

This is pretty hard to test since this test and compiler and everything has been the same since forever, and this is the first time I've seen this issue. (It also doesn't happen to me locally, and tsan doesn't complain.)

Zer0-One (approved):

